PR #36, opened 2018-09-15
[WIP] consolidate ci tests

jamiecook:
Making some progress.. still to do
  - [ ] figure out what needs to be published from each platform (just the AoN.so file?)
  - [ ] update ci.sh to package those files
  - [ ] update yml to artifact those files
  - [ ] figure out how to package the binary files with the other py files for release to pip (probably just do this from one of the platforms - suggest linux)
  - [ ] figure out how to publish to pip 

image
image

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -14,7 +14,7 @@ main() {
         package_for_publication
     else
         echo "Unknown option ${1}"
-        exit(-1)
+        exit -1
     fi
 
     run_tests
--- a/ci.sh
+++ b/ci.sh
@@ -10,7 +10,7 @@ main() {
 
     if [[ "$1" == "test" ]]; then
         run_tests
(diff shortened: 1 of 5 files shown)


ci.sh
```diff
@@ -0,0 +1,39 @@
...
+    run_tests
+}
+
+setup_pip() {
+    python -m pip install --upgrade pip
+}
+
+install_requirements() {
+    pip install -r requirements.txt
+    pip install numpy --upgrade
```
jamiecook: @pedrocamargo ... why are you upgrading numpy here?
requirements.txt already has it included right?

pedrocamargo: It should have the latest

jamiecook:
@pedrocamargo ... made some progress - need to have face to face chat with you about the remainder

PR #45, opened 2018-12-27
Replaces xrange with range in Cython code

pedrocamargo:
Not a particular requirement, but helps in automatic code inspections


setup.py
```diff
@@ -13,7 +13,7 @@
 ext_module = Extension('AoN',
                        [whole_path],
                        include_dirs=[np.get_include()])
-if __name__ == "__main__":
+if __name__ == "_  _main__":
```
jamiecook: Extra space?

pedrocamargo: yup. Good catch!  Not sure how it passed...

PR #59, opened 2019-06-02
Adds robust import of OMX for matrix and index names that are too long

pedrocamargo:


aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -248,20 +271,33 @@ def create_from_omx(
             raise ValueError("AequilibraE only supports square matrices")
         zones = shp[0]
 
+        if robust:
+            core_names = []
+            for c in do_cores:
```
jamiecook: Can't this just be a comprehension like the else case?

pedrocamargo: No.  core_names is an input to the computation of its next element

jamiecook: ah

```python
core_names = functools.reduce(lambda acc,n: robust_name(n, CORE_NAME_MAX_LENGTH, acc),do_cores, [])
idx_names = functools.reduce(lambda acc,n: robust_name(n, INDEX_NAME_MAX_LENGTH, acc),do_idx, [])
```

pedrocamargo: Incorporated!  Thanks!

PR #64, opened 2019-09-13
Pedro/omx support

pedrocamargo:


pedrocamargo:
The idea of this pull request is to allow users to use OMX matrices as AequilibraE ones. It is read-only, but all the methods used to read data are adapted to accept OMX, and the others raise errors for no implementation

pedrocamargo:
Documentation has not been updated, however.

pedrocamargo:
Can @jamiecook or @ychtw review this?

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -5,9 +5,18 @@
 from shutil import copyfile
 from typing import List
 import functools
+import logging
 
 import numpy as np
 
+# Checks if we can display OMX
+spec = iutil.find_spec("openmatrix")
+has_omx = False
```
jamiecook:
```suggestion
has_omx = spec is not None
if has_omx:
    import openmatrix as omx
```

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -171,7 +182,9 @@ def create_empty(
                                 "than {}: {}".format(INDEX_NAME_MAX_LENGTH, ind_name)
                             )
                     else:
```
jamiecook: Try moving your raise to the true case... then you don't need an else case at all

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -218,7 +233,9 @@ def create_from_omx(
         :return:
         """
 
-        def robust_name(input_name: str, max_length: int, forbiden_names: List[str]) -> str:
+        def robust_name(
```
jamiecook: black?

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -254,7 +269,9 @@ def robust_name(input_name: str, max_length: int, forbiden_names: List[str]) ->
             do_cores = [x for x in cores if x in avail_cores]
             if cores != do_cores:
                 do_cores = [x for x in cores if x not in avail_cores]
```
jamiecook: isn't this "missing_cores" rather than "do_cores"

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -352,93 +399,199 @@ def __load__(self):
...
+        self.num_indices = len(self.index_names)
+
+        # Zones
+        self.zones = len(list(self.omx_file.mapping(self.index_names[0]).keys()))
+
+        # Matrix cores
+        self.cores = len(self.names)
+
+        # Index
+        self.set_index(self.index_names[0])
```
jamiecook: Not sure how much these comments add

pedrocamargo: Agreed

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: Done

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -293,41 +326,57 @@ def robust_name(input_name: str, max_length: int, forbiden_names: List[str]) ->
...
-            self.index[:] = np.arange(zones)
+            self.index[:, 0] = np.arange(zones)
 
         self.indices.flush()
 
-    def __load__(self):
+    def __load_aem__(self):
         # GET File version
-        self.__version__ = np.memmap(self.file_path, dtype="uint8", offset=0, mode="r+", shape=1)[0]
+        self.__version__ = np.memmap(
```
jamiecook: I'd clean this up so that it all fits on one line and makes it easy to see the overall flow of the binary format

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: It happened in a lot of places. I think I corrected most of them.

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: Nope.It is correct. We fix the names if needed

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: Done

PR #82, opened 2020-01-05
Version 0.6

pedrocamargo:
Brings the AequilibraE project to life


pedrocamargo:
@jamiecook , the problem with the tests with GitHub Actions is that their distirbution of Python-sqlite does not support loading extensions.  Do you have any experience with using Docker on GitHub so we can get around this?

Travis is also acting up due to memory errors, so not very reliable atm

jamiecook:
Hey @pedrocamargo - I haven't used github actions at all.... i'd say that's it's probably using docker under the hood anyway though. 

You might try out the suggestions in this stackoverflow question about using pysqlite2 
to get a more up to date version of sqlite

https://stackoverflow.com/questions/1545479/force-python-to-forego-native-sqlite3-and-use-the-installed-latest-sqlite3-ver

You might be able to add steps to your tests.yaml (where the dependencies are installed) to also easy_install the correctly compiled version.

pedrocamargo:
@jamie,  pysqlite has been deprecated and merged into the Python core library, but the containers that Github action uses are loaded with a version of Python that does not allow sqlite to load extensions.

pedrocamargo:
Can @jamiecook , @arobrien or @ychtw   review this one?

aequilibrae/parameters.yml
```diff
@@ -17,3 +17,187 @@ system:
   logging: true
   report zeros: 1
   temp directory: /temp
+  logging_directory: /temp
+  spatialite_path: C:\Users\pedro\Documents\mod_spatialite-NG-win-amd64
```
ychtw: should be C:\mod_spatialite-NG-win-amd64 or other path that is not user-specific?

pedrocamargo: This is in the parameters file, which is user-defined. This is just the example.

There is documentation on it:  https://aequilibrae.com/python/dev/gettingstarted.html#spatialite

pedrocamargo:
@janzill , can you take a crack a reviewing this?

janzill:
> @janzill , can you take a crack a reviewing this?

Will do on the weekend

pedrocamargo:
Awesome, @janzill !  Thanks!!!!

aequilibrae/project/network/network.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self, project):
+        WorkerThread.__init__(self, None)
+
+        self.conn = project.conn  # type: sqlc
+
+    def _check_if_exists(self):
+        curr = self.conn.cursor()
+        curr.execute("SELECT count(*) FROM sqlite_master WHERE type='table' AND name='links';")
+        tbls = curr.fetchone()[0]
+        if tbls > 0:
```
jamiecook:
```suggestion
        return table_count > 0
```

aequilibrae/project/network/network.py
```diff
@@ -0,0 +1,198 @@
...
+            return True
+        return False
```
jamiecook:
```suggestion
return table_count > 0

File: aequilibrae/project/network/network.py [hunk 4]
```diff
@@ -0,0 +1,198 @@
...
+
+        sql = """CREATE TABLE 'links' (
+                          ogc_fid INTEGER PRIMARY KEY,
+                          link_id INTEGER UNIQUE NOT NULL,
+                          {});"""
+
+        flds = fields["one-way"]
+
+        owlf = [
+            "{} {}".format(list(f.keys())[0], f[list(f.keys())[0]]["type"])
```
jamiecook: I feel like this pattern is in use a lot - think about a method
```python
def first_key(d):
    return list(d.keys())[0]

Comment (jamiecook) [reply aequilibrae/project/network/network.py]: and then maybe think about why you need to do this so often - isn't there a better structure for storing this data?

File: aequilibrae/project/network/network.py [hunk 5]
```diff
@@ -0,0 +1,198 @@
...
+            for f in flds
+            if list(f.keys())[0].upper() != "LINK_ID"
+        ]
+
+        flds = fields["two-way"]
+        twlf = []
+        for f in flds:
+            nm = list(f.keys())[0]
+            tp = f[nm]["type"]
+            twlf.extend(["{}_ab {}".format(nm, tp), "{}_ba {}".format(nm, tp)])
```
jamiecook: Common py3.6+

https://realpython.com/python-f-strings/

aequilibrae/project/network/network.py
```diff
@@ -0,0 +1,198 @@
...
+    def add_network_triggers(self) -> None:
+        curr = self.conn.cursor()
+        logger.info("Adding data indices")
+
+        curr.execute("""CREATE INDEX links_a_node_idx ON links (a_node);""")
+        curr.execute("""CREATE INDEX links_b_node_idx ON links (b_node);""")
+
+        pth = os.path.dirname(os.path.realpath(__file__))
+        qry_file = os.path.join(pth, "network_triggers.sql")
+        sql_file = open(qry_file, "r")
```
jamiecook:
```
with open(qry_file,'r') as f:

File: aequilibrae/project/network/osm_builder.py [hunk 7]
```diff
@@ -0,0 +1,289 @@
...
+        del alinks
+
+        logger.info("Finalizing data structures")
+        self.building.emit(["text", "Finalizing data structures"])
+
+        node_count = self.unique_count(np.array(all_nodes))
+
+        return node_count
+
+    def importing_links(self, node_count):
```
jamiecook: this method feels pretty long ... can it be broken up at all?

tests/aequilibrae/project/test_network.py
```diff
@@ -0,0 +1,102 @@
...
+        pth = os.getcwd()
+        if "WINDOWS" in plat.upper():
+            par = Parameters()
+            spatialite_path = par.parameters["system"]["spatialite_path"]
+            if os.path.isfile(os.path.join(spatialite_path, "mod_spatialite.dll")):
+                os.chdir(spatialite_path)
+        try:
+            self.conn.load_extension("mod_spatialite")
+        except Exception as e:
+            warn("AequilibraE might not work as intended without spatialite. {}".format(e.args))
```
jamiecook: You might consider pulling this out to a method

```python
def get_spatialite_connection(file_path):

File: tests/aequilibrae/project/test_osm_downloader.py [hunk 9]
```diff
@@ -0,0 +1,25 @@
...
+
+    def test_do_work2(self):
+        # LITTLE PLACE IN THE MIDDLE OF THE Grand Canyon north rim
+        self.o = OSMDownloader([[-112.185, 36.59, -112.179, 36.60]], ["car"])
+        self.o.doWork()
+        if len(self.o.json[0]["elements"]) > 1000:
+            self.fail("It found too many elements in the middle of the Grand Canyon")
+
+        if len(self.o.json[0]["elements"]) < 10:
+            self.fail("It found too few elements in the middle of the Grand Canyon")
```
jamiecook: Are you sure you want to automatically run osm downloading in your automated test suite?

tests/aequilibrae/project/test_project.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class TestProject(TestCase):
+    def tearDown(self) -> None:
+        if os.path.isfile(temp_proj_name):
+            os.unlink(temp_proj_name)
+
+    def test_creation(self):
+        test_file = temp_proj_name
+        with self.assertRaises(FileNotFoundError):
```
jamiecook: you can also use

```
self.assertRaises(FileNotFoundError, Project, test_file)

tests/aequilibrae/project/test_osm_downloader.py
pedrocamargo: That is the only way I could find to really test this.  I chose a little patch in the middle of the Grand Canyon National part that should be very stable.

aequilibrae/project/network/network.py
pedrocamargo: Wow!! I didn't know this one!!

tests/aequilibrae/project/test_osm_downloader.py
jamiecook: I was actually meaning the etiquette of hitting a non-profit's public API over and over. 

I'd consider getting it working and then marking as 'on-hold', or putting a "only runs one time in 10" in the test

aequilibrae/project/network/network.py
jamiecook: https://media0.giphy.com/media/xT0xeJpnrWC4XWblEk/giphy.gif

pedrocamargo: I also didn't know we could use GIFs here.... Hahahaha

tests/aequilibrae/project/test_osm_downloader.py
pedrocamargo: Oh. I see.  Done.

pedrocamargo: I am making once every 20 times

aequilibrae/project/network/network.py
pedrocamargo: I went ahead and changed it in a lot of places.  Might have a future PR just to change that

tests/aequilibrae/project/test_project.py
pedrocamargo: applied here, but I would need to change this in many other places.  Will try to learn this pattern and use it in the future

aequilibrae/project/network/network.py
pedrocamargo: Excellent suggestion. done

aequilibrae/project/network/network.py
jamiecook: just note that f-strings are 3.6+

aequilibrae/project/network/network.py
pedrocamargo: Not really.  The problem here is that the data is all organized as dictionary of dictionaries, so we are driving down sub-dictionaries

aequilibrae/project/network/osm_builder.py
pedrocamargo: I break it up in the next version 0.6.1

pedrocamargo:
Thanks for the help, @jamiecook !!!

PR #92, opened 2020-02-08
[WIP] Jan/more descent directions

janzill:
I've tested conjugate and biconjugate Frank-Wolfe assignment on single vehicle class instances, multiple vehicle classes needs testing (and probably fixing). The calculation of the descent direction is very slow on larger networks (like Chicago regional), we will have to port this to cython if we want decent runtimes on real-world networks.


janzill:
@pedrocamargo I was thinking about the structure and the amount of code duplication, we could either go with what we talked about (base class EquilibriumAssigner and then deriving from it to overwrite step direction finder) or simply use the current BFW code and then add a little bit of logic to set some flags because MSA, FW, CFW, and BFW are basically implemented here already.

janzill:
The only disadvantage in case of the latter is that we would allocate one more load object in memory than we would have to because we need to keep the previous two step directions around, but that's a minor thing that won't impact performance in practical situations. Any preferences on your side? I'm leaning towards the second option, but that's purely personal preference.

janzill:
I went ahead with creating one class named LinearApproximation, which is identical to the previous BFW class, except that the step direction calculation now checks for the algorithm name. Also, the step size finder exists early with 1/iteration_number if msa has been specified. Minimal code changes, but we now have everything in one place and no more code duplication

aequilibrae/paths/linear_approximation.py
```diff
@@ -0,0 +1,308 @@
...
+from aequilibrae import logger
+
+if False:
+    from aequilibrae.paths.traffic_assignment import TrafficAssignment
+
+
+class LinearApproximation:
+    def __init__(self, assig_spec, algorithm) -> None:
+        self.algorithm = algorithm
+        parameters = Parameters().parameters["assignment"]["equilibrium"]
```
jamiecook: Is this a global variable?

pedrocamargo: There is a parameter file that ships with AequilibraE.
Documentation still incomplete:  https://aequilibrae.com/python/dev/parameter_file.html
But there is a little example on how to manipulate it: https://aequilibrae.com/python/dev/usageexamples.html#parameters-module

pedrocamargo:
@janzill,  This now passes all tests (it is missing traffic assignment tests, but I am still preparing instances that can be used for such).

What do you think if we merge it into the V0.6.1 upcoming release branch?

janzill:
Sounds good :)

PR #98, opened 2020-02-19
QGIS changes

pedrocamargo:
This pull request contains one small bug fix reported after release 0.6.1 and many changes necessary for the integration with the QGIS GUI


aequilibrae/project/network/osm_builder.py
```diff
@@ -111,14 +120,17 @@ def importing_links(self, node_count):
...
+        self.__emit_all(["maxValue", L])
 
         nodes_to_add = set()
         counter = 0
         mode_codes, not_found_tags = self.modes_per_link_type()
 
         for osm_id, link in self.links.items():
             self.__emit_all(["Value", counter])
+            counter += 1
+            self.__emit_all(["text", f"{counter:,} of {L:,} super links added"])
```
janzill: maybe redundant because you have error logging, but this is "about to add" or "adding", not added, right?

setup.py
```diff
@@ -40,9 +40,9 @@
         license="See license.txt",
         classifiers=[
             "Programming Language :: Python",
-            "Programming Language :: Python :: 3.5",
             "Programming Language :: Python :: 3.6",
             "Programming Language :: Python :: 3.7",
+            "Programming Language :: Python :: 3.8",
```
janzill: always forward :)

aequilibrae/project/network/network.py
```diff
@@ -52,7 +54,17 @@ def skimmable_fields(self):
                     break
 
         all_fields.append('distance')
-        return all_fields
+        real_fields = []
```
janzill: I'm not entirely sure what is happening here, do you mind reminding me of the data structure?

aequilibrae/paths/linear_approximation.py
```diff
@@ -331,10 +358,19 @@ def derivative_of_objective(stepsize):
...
+                min_res = root_scalar(derivative_of_objective, bracket=[0, 1])
+                self.stepsize = min_res.root
+                if not min_res.converged:
+                    logger.warn("Descent direction stepsize finder is not converged")
+            else:
+                min_res = root_scalar(derivative_of_objective, 1 / self.iter)
+                if not min_res.success:
+                    logger.warn("Descent direction stepsize finder is not converged")
+                self.stepsize = min_res.x[0]
+                if self.stepsize <= 0.0 or self.stepsize >= 1.0:
```
janzill: this should be <0 and >1, ==0 and ==1 are handled as edge cases below

aequilibrae/project/network/network.py
pedrocamargo: What this does is to return all fields that can be skimmed. Default fields (distance excepted) should not be considered skimmable, so what we are doing here is to get all numeric fields except the required fields, plus distance

aequilibrae/project/network/osm_builder.py
pedrocamargo: True

aequilibrae/paths/linear_approximation.py
pedrocamargo: This is the only way I got things to work with this older version of SciPy.  Results we get from them are a bit weird (and quite a bit slower too)

aequilibrae/project/network/network.py
janzill: right, thanks

aequilibrae/paths/linear_approximation.py
janzill: wait, it's wrapped in a try block so it'll do the right thing when it catches the exception, sorry I missed that bit of logic, all good

aequilibrae/paths/linear_approximation.py
```diff
@@ -10,12 +11,32 @@
 from aequilibrae.paths.AoN import copy_one_dimension, copy_two_dimensions, copy_three_dimensions
 from aequilibrae import logger
 
+try:
```
jamiecook: why not just use scipy.__version__ ??

image

aequilibrae/project/network/network.py
```diff
@@ -256,12 +268,17 @@ def set_time_field(self, time_field: str) -> None:
 
     def count_links(self) -> int:
         c = self.conn.cursor()
-        c.execute("""select count(*) from links""")
+        c.execute("""select count(link_id) from links""")
```
jamiecook: These could all be count(*) with a where condition

```python
    def count(tablename, where_condition=""):
        c = self.conn.cursor()
        return c.execute(f"select count(*) from {tablename} {where_condition};").fetchone()[0]

File: aequilibrae/project/spatialite_connection.py [hunk 7]
```diff
@@ -0,0 +1,21 @@
...
+
+def spatialite_connection(conn):
+    conn.enable_load_extension(True)
+    plat = platform.platform()
+    pth = os.getcwd()
+    if "WINDOWS" in plat.upper():
+        par = Parameters()
+        spatialite_path = par.parameters["system"]["spatialite_path"]
+        if os.path.isfile(os.path.join(spatialite_path, "mod_spatialite.dll")):
+            os.chdir(spatialite_path)
```
jamiecook: here you can just add that dir to the path 

image

pedrocamargo: It did not work on windows

jamiecook: note the difference in the sys.path vs os.environ["Path"]. sys.path isn't enough

aequilibrae/project/network/network.py
pedrocamargo: The count(*) worked fine too, but I wanted to make it a little more readable. That's all

aequilibrae/project/spatialite_connection.py
pedrocamargo: Hummm...  Good point

pedrocamargo: And it worked!  Will add this to a future PR

aequilibrae/project/network/network.py
pedrocamargo: Oh. Now I see the point. A single method for all of them...

aequilibrae/paths/linear_approximation.py
pedrocamargo: That was me trying to get around Flake8. But that works as well (did not think it would)

aequilibrae/project/network/network.py
jamiecook: #noqa for the win in these situations

aequilibrae/paths/linear_approximation.py
jamiecook: #noqa for the win in these situations

Issue #103, opened 2020-02-26
Skims are wrong


pedrocamargo:
It seems that when we generate skims for more than one field (say blended distance and final travel time), we are getting the labels as we want, but all the values for one of them only

pedrocamargo:
Commit bfe8985 had the wrong message.  that was fixing issue #102 , actually.

pedrocamargo:
The error has been found. 
The issue is that we do not currently update the skimming array in the graph (which is what we use to skim) when we recompute the time after each iteration

Issue #102, opened 2020-02-26
Skims not generated for zones with zero demand


pedrocamargo:
Skims are just not being generated for zones with no demand.

pedrocamargo:
Fixed on new branch

PR #104, opened 2020-02-26
Skim issues

pedrocamargo:
I detected two issues with skim generation, reported as issues #102 and #103 .  This pull request fixes both of them


janzill:
noice. should probably add more tests, which is a generic "todo"

Issue #105, opened 2020-02-26
Synthetic gravity model calibration fails when using OMX inputs


pedrocamargo:
Computation fails hard, and there are a few issues:

* Impedance matrix gets copied into a new array, but copying OMX files is implemented in a too lazy of a way
* Gravity application procedure does not leverage computational_view, attempting to access the dictionary of matrices

PR #106, opened 2020-02-26
Trip distribution now works with OMX matrices

pedrocamargo:
Fixes issue #105


pedrocamargo:
@janzill , I am assigning reviews to you because you are just too efficient.  If you  are not in the mood of doing, or just can't do it, feel free to do like @arobrien and @jamiecook and just ignore me!

aequilibrae/paths/linear_approximation.py
```diff
@@ -10,16 +10,14 @@
 from aequilibrae.paths.AoN import triple_linear_combination, triple_linear_combination_skims
 from aequilibrae.paths.AoN import copy_one_dimension, copy_two_dimensions, copy_three_dimensions
 from aequilibrae import logger
-
-try:
+import scipy
+if int(scipy.__version__.split('.')[1]) >= 3:
```
jamiecook: You could also consider using ```semver``` to avoid edge cases

image

This code would flag scipy version 2.1 as old - ```if semver.compare(semver.parse_version_info(scipy.__version__), "1.4.0") < 0```

aequilibrae/starts_logging.py
```diff
@@ -34,4 +35,14 @@ def StartsLogging():
     return logger
 
 
+def cleaning():
+    p = tempfile.gettempdir() + "/aequilibrae_*"
+    for f in glob.glob(p):
+        try:
+            os.unlink(f)
```
jamiecook: you can also just use pathlib and let it do the error ignoring, have i mentioned how much I hate exceptions

pedrocamargo: I also hate exceptions, but there is no way to check if a file is in use somewhere else in the system before attempting to remove it

aequilibrae/paths/linear_approximation.py
pedrocamargo: I looked into similar stuff, but there is nothing like that available in the Python distribution that comes with QGIS for Windows, so I am a bit stuck.  I actually hope to deprecate the use of SciPy older than 1.3 as soon as the QGIS Windows distribution gets updated.

PR #109, opened 2020-03-04
API documentation

pedrocamargo:
Complete API documentation and some small changes to the API that became obvious when writing the documentation.

Built documentation can be seen on www.aequilibrae.com/python/dev

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -35,9 +35,14 @@ Contents
    qgis
 
 
-Indices and tables
-==================
+Older versions
+==============
 
-* :ref:`genindex`
-* :ref:`modindex`
-* :ref:`search`
+AequilibraE has been evolving quite fast, so we recommend you upgrading to a
+newer version as soon as you can.  In the meantime, you can find the
+documentation for older versions below.
+
+* `0.5.3 <https://aequilibrae.com/python/V.0.5.3/>`_
+* `0.6.0 <https://aequilibrae.com/python/V.0.6.0/>`_
+* `0.6.1 <https://aequilibrae.com/python/V.0.6.1/>`_
+* `0.6.2 <https://aequilibrae.com/python/V.0.6.2/>`_


.gitignore
```diff
@@ -13,7 +13,7 @@ _templates
 .idea/misc.xml
 .idea/modules.xml
 .idea/vcs.xml
-
+.idea/*
```
janzill: I don't know about this one, the .idea folder is a whole discussion to itself

aequilibrae/distribution/gravity_calibration.py
```diff
@@ -31,15 +14,68 @@
 
 
 class GravityCalibration:
-    """"
-        where function is: 'EXPO' or 'POWER'. 'GAMMA' and 'FRICTION FACTORS' to be implemented at a later time
-        parameters are: 'max trip length'
+    """
+        Calibrate a traditional gravity model
+
+        Available deterrence function forms are: 'EXPO' or 'POWER'. 'GAMMA'
```
janzill: are these three options?

aequilibrae/paths/all_or_nothing.py
```diff
@@ -7,21 +7,25 @@
...
-from .results import AssignmentResults
-from .graph import Graph
 from aequilibrae.matrix import AequilibraeMatrix
 
 spec = iutil.find_spec("PyQt5")
 pyqt = spec is not None
 if pyqt:
     from PyQt5.QtCore import pyqtSignal as SIGNAL
 
+if False:
```
janzill: you don't need these anymore?

aequilibrae/paths/parameters.pxi
```diff
@@ -15,6 +15,6 @@ cdef ITYPE_t NULL_IDX = 18446744073709551615
 cdef double INFINITE = 1.79769313e+308
 
 VERSION = 0.6
-MINOR_VRSN = 2
+MINOR_VRSN = 3
 binary_version = "0.6.1"
```
janzill: is the binary version different from ```str(VERSION) + "." + str(MINOR_VRSN)```

aequilibrae/paths/results/assignment_results.py
```diff
@@ -193,7 +212,14 @@ def setSavePathFile(self, save=False, path_result=None):
 
         self.path_file = {"save": save, "results": a}
 
-    def get_load_results(self):
+    def get_load_results(self) -> AequilibraeData:
+        """
+        Resulting number of cores will be adjusted to a minimum of zero or the maximum available in the system if the
```
janzill: I don't think you copy-pasted the right text here :D

tests/aequilibrae/distribution/test_gravityApplication.py
```diff
@@ -1,5 +1,6 @@
 from unittest import TestCase
-from aequilibrae.matrix import AequilibraeData, AequilibraeMatrix
+from aequilibrae.matrix import AequilibraeData
```
janzill: same, same

aequilibrae/distribution/gravity_calibration.py
pedrocamargo: Yes

aequilibrae/paths/all_or_nothing.py
pedrocamargo: I do, but just for Typing

aequilibrae/paths/parameters.pxi
pedrocamargo: Yes. This is because of AequilibraE for QGIS, which gets the Python package delivered with it and downloads the binaries manually.

tests/aequilibrae/distribution/test_gravityApplication.py
pedrocamargo: I was chasing a bug here...

Issue #84, opened 2020-01-06
Wheel files for WINDOWS are not being produced automatically


pedrocamargo:
It is not clear why it is not working. Need to investigate

PR #111, opened 2020-03-09
Importing structure for building

pedrocamargo:
Building AequilibrE was not possible without prior compilation of Cython code, which is not ideal.  So I change some of the imports throughout the software
Also added new workflows for testing under Windows and building Windows packages for 32 abd 64 bits

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -25,13 +25,6 @@ jobs:
         choco install visualstudio2019community
         choco install -y microsoft-build-tools visualcppbuildtools --version 14.0.25420.1
 
-    - name: Compile library
-      run: |
-        cd aequilibrae/paths
-        python setup_Assignment.py build_ext --inplace
-        cd ..
-        cd ..
-
     - name: Build
       run: python setup.py sdist bdist_wheel
 
--- a/tests/aequilibrae/matrix/test_aequilibraeMatrix.py
+++ b/tests/aequilibrae/matrix/test_aequilibraeMatrix.py
@@ -3,27 +3,29 @@ import os
 import random
 import openmatrix as omx
 from unittest import TestCase
-
+from os.path import dirname, join
 import numpy as np
-
+import uuid
 from aequilibrae.matrix import AequilibraeMatrix
 from ...data import omx_example, no_index_omx, siouxfalls_skims
 
 zones = 50
-name_test = AequilibraeMatrix().random_name()
-copy_matrix_name = AequilibraeMatrix().random_name()
-csv_export_name = copy_matrix_name + ".csv"
-omx_export_name = copy_matrix_name + ".omx"
 
 
 class TestAequilibraeMatrix(TestCase):
     matrix = None
 
     def setUp(self) -> None:
+        temp_folder = dirname(dirname(dirname(os.path.abspath(__file__))))
+        self.name_test = temp_folder + f"/Aequilibrae_matrix_{uuid.uuid4()}.aem"
+        self.copy_matrix_name = temp_folder + f"/Aequilibrae_matrix_{uuid.uuid4()}.aem"
+        self.csv_export_name = temp_folder + f"/Aequilibrae_matrix_{uuid.uuid4()}.csv"
+        self.omx_export_name = temp_folder + f"/Aequilibrae_matrix_{uuid.uuid4()}.omx"
+
         if self.matrix is not None:
             return
         args = {
-            "file_name": name_test,
+            "file_name": self.name_test,
             "zones": zones,
             "matrix_names": ["mat", "seed", "dist"],
             "index_names": ["my indices"],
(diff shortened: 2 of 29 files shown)


tests/aequilibrae/matrix/test_aequilibraeMatrix.py
```diff
@@ -43,8 +45,11 @@ def setUp(self) -> None:
 
     def tearDown(self) -> None:
         try:
-            os.remove(name_test) if os.path.exists(name_test) else None
-            self.matrix = None
+            del self.matrix
```
janzill: Creating a temp directory within tests and then deleting it at the end will save you from having to delete them individually

.github/workflows/build_windows.yml
```diff
@@ -11,19 +11,19 @@ jobs:
       matrix:
         python-version: [ '3.6', '3.7', '3.8' ]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
janzill: you're using v1 everywhere else, right? Is this an upgrade step-by-step, or are some workflows incompatible with v2?

tests/aequilibrae/project/test_network.py
```diff
@@ -7,31 +7,25 @@
...
 from aequilibrae.parameters import Parameters
+from os.path import join, dirname
 from warnings import warn
 from random import random
+from aequilibrae.project.spatialite_connection import spatialite_connection
 
 
 class TestNetwork(TestCase):
     def setUp(self) -> None:
+        spatialite_folder = dirname(dirname(dirname(dirname(os.path.abspath(__file__)))))
```
janzill: I wonder if there's a way to do this nicer, it's used in all of the tests and I don't like it. Having it in one place would also make it easier to change.

tests/aequilibrae/matrix/test_aequilibraeMatrix.py
jamiecook: yeah, you can use ```with tempfile.tempDirectory() as temp_dir:``` to do it all in one neat block

tests/aequilibrae/project/test_network.py
jamiecook: in rfmd i put everything into a spatialite_utils.py file that had an ```ensure_spatialite_on_path``` function that would add it if it didn't exist

jamiecook:
```python
def add_mod_spatialite_path():
    """Ensure that the spatialite dll is on the path."""
    root_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
    spatialite_dir = os.path.join(root_dir, "bin", "mod_spatialite-4.3.0a-win-amd64")
    if spatialite_dir not in os.environ["PATH"]:
        os.environ["PATH"] = spatialite_dir + ";" + os.environ["PATH"]
```

tests/aequilibrae/project/test_network.py
```diff
@@ -7,31 +7,25 @@
...
-            par = Parameters()
-            spatialite_path = par.parameters["system"]["spatialite_path"]
-            if os.path.isfile(os.path.join(spatialite_path, "mod_spatialite.dll")):
-                os.chdir(spatialite_path)
-        try:
-            self.conn.load_extension("mod_spatialite")
-        except Exception as e:
-            warn(f"AequilibraE might not work as intended without spatialite. {e.args}")
-        os.chdir(pth)
+        self.conn = spatialite_connection(self.conn)
```
jamiecook: you could also pull this functionality there ```spatial_connect(self.file2)```

tests/setup_windows_spatialite.py
```diff
@@ -0,0 +1,15 @@
+import urllib.request
+import zipfile
+from os.path import join, dirname
+import os
+
+pth = 'https://github.com/AequilibraE/aequilibrae/releases/download/V0.6.0.post1/mod_spatialite-NG-win-amd64.zip'
```
jamiecook: where did you get this zip from?
I had real problems with the version i found - I had to replace the c++ std runtime dll files with up to date versions :/
https://github.com/sqlitebrowser/sqlitebrowser/wiki/SpatiaLite-on-Windows 
Did you have any such problems?

tests/aequilibrae/matrix/test_aequilibraeMatrix.py
pedrocamargo: That is failing in GitHub Actions under Windows....  I tried...  I am deleting the files to not pollute my own system when I am running the tests locally...

.github/workflows/build_windows.yml
pedrocamargo: I need to upgrade everything to 2.0

tests/aequilibrae/project/test_network.py
pedrocamargo: The problem is that the path to Spatialite is not always the same. When running the tests on GitHub, Spatialite needs to be downloaded. I changed the way it happens so it is a little more elegant

.github/workflows/build_windows.yml
pedrocamargo: Done

tests/aequilibrae/project/test_network.py
pedrocamargo: Done.

tests/setup_windows_spatialite.py
pedrocamargo: I got it directly from here.  http://www.gaia-gis.it/gaia-sins/windows-bin-NEXTGEN-amd64/

Just uploaded to the repository as a way to make sure all tests work in case the original url fails (It is linked in the documentation)

pedrocamargo:
Fixes issue #84

PR #122, opened 2020-03-30
Update depreciated warn method calls

alexdawn:
logger.warn() has been depreciated since python 3.3. This replaces
logger.warn() with logger.warning()


pedrocamargo:
Thanks for the contributions, @alexdawn

Issue #124, opened 2020-04-01
Both 'cfw', 'bfw' and get stuck with zero step size


alexdawn:
With version 0.6.3

Both Both 'cfw', 'bfw' and get stuck with zero step size while MSA and Frank-Wolfe options work. The log for BFW is shown below:

```
2020-04-01 11:49:53,085;aequilibrae;INFO ; bfw Assignment STATS
2020-04-01 11:49:53,089;aequilibrae;INFO ; Iteration, RelativeGap, stepsize
2020-04-01 11:49:53,089;aequilibrae;INFO ; Iteration, RelativeGap, stepsize
2020-04-01 11:49:53,207;aequilibrae;INFO ; 1,inf,1.0
2020-04-01 11:49:53,207;aequilibrae;INFO ; 1,inf,1.0
2020-04-01 11:49:53,325;aequilibrae;INFO ; 2,0.01813536336911382,0.2561015932062774
2020-04-01 11:49:53,325;aequilibrae;INFO ; 2,0.01813536336911382,0.2561015932062774
2020-04-01 11:49:53,441;aequilibrae;INFO ; 3,0.005430047866020795,0.0
2020-04-01 11:49:53,441;aequilibrae;INFO ; 3,0.005430047866020795,0.0
2020-04-01 11:49:53,561;aequilibrae;INFO ; 4,0.005430047866020795,0.0
2020-04-01 11:49:53,561;aequilibrae;INFO ; 4,0.005430047866020795,0.0
2020-04-01 11:49:53,676;aequilibrae;INFO ; 5,0.005430047866020795,0.0
2020-04-01 11:49:53,676;aequilibrae;INFO ; 5,0.005430047866020795,0.0
2020-04-01 11:49:53,794;aequilibrae;INFO ; 6,0.005430047866020795,0.0
2020-04-01 11:49:53,794;aequilibrae;INFO ; 6,0.005430047866020795,0.0
....
```

and so on until it hits the max iteration count.
The graph and demand was constructed using the Anaheim model and the assignment code copying the Sioux Falls example. I've checked the `graph.network` field names are correct.

```python

graph.set_graph('time') # let's say we want to minimize time
graph.set_skimming(['time', 'distance']) # And will skim time and distance
graph.set_blocked_centroid_flows(True)

demand = aem
# demand.load(os.path.join(data_folder, 'Anaheim_trips.omx'))
[...]

PR #125, opened 2020-04-03
Workflows & bug fix

pedrocamargo:
* It updates the Github action flows to upload the sphinx documentation to the new hosting solution (AWS S3)

* Address issue #124

### DIFF
--- a/aequilibrae/paths/linear_approximation.py
+++ b/aequilibrae/paths/linear_approximation.py
@@ -387,7 +387,7 @@ class LinearApproximation(WorkerThread):
             # in order to add a small fraction of the AoN. A heuristic value equal to the corresponding MSA step size
             # seems to work well in practice.
             if derivative_of_objective(0.0) < derivative_of_objective(1.0):
-                if self.algorithm == "frank-wolfe":
+                if self.algorithm == "frank-wolfe" or self.stepsize == 0.0:
                     heuristic_stepsize_at_zero = 1.0 / self.iter
                     logger.warn("# Alert: Adding {} to stepsize to make it non-zero".format(heuristic_stepsize_at_zero))
                     self.stepsize = heuristic_stepsize_at_zero
--- a/aequilibrae/paths/linear_approximation.py
+++ b/aequilibrae/paths/linear_approximation.py
@@ -13,7 +13,7 @@ try:
     from aequilibrae.paths.AoN import triple_linear_combination, triple_linear_combination_skims
     from aequilibrae.paths.AoN import copy_one_dimension, copy_two_dimensions, copy_three_dimensions
 except ImportError as ie:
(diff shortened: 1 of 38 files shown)


aequilibrae/paths/linear_approximation.py
```diff
@@ -397,6 +397,10 @@ def derivative_of_objective(stepsize):
                     self.stepsize = 0.0
                     # need to reset conjugate / bi-conjugate direction search
                     self.do_fw_step = True
+
+                    # By doing it recursively, we avoid doing the same AoN again
```
janzill, line 401: I took me a second to understand this, but you are right, no need to re-do the AON, although the logic is a bit more complicated now.

aequilibrae/paths/linear_approximation.py
```diff
@@ -387,7 +387,7 @@ def derivative_of_objective(stepsize):
             # in order to add a small fraction of the AoN. A heuristic value equal to the corresponding MSA step size
             # seems to work well in practice.
             if derivative_of_objective(0.0) < derivative_of_objective(1.0):
-                if self.algorithm == "frank-wolfe":
+                if self.algorithm == "frank-wolfe" or self.stepsize == 0.0:
```
janzill, line 390: I don't think we need to check for zero here because of the preceding line

pedrocamargo: This is EXACTLY what fixes the issue. When the algorithm is NOT FW, but we found a stepsize of zero

aequilibrae/paths/linear_approximation.py
pedrocamargo: Indeed

aequilibrae/paths/linear_approximation.py
janzill: I am not sure we should special case it, this will make a CFW or BFW step with zero contribution add a little bit (see my TODO on line 396 below). It might be better to just re-start from FW. Have you tested this?

pedrocamargo: Tested what I implemented?  I did it with the exact instance @alexdawn tested with.

What is going on here (for CFW/BFW) is the following:

If the step size is zero (generally when we fail to find a step size):
    Do a FW step:
        if we fail to do compute a step size (we find ourselves in this part of the code by either being FW or due to having a do_fw_step true:
                do MSA step

Does it make sense?

janzill: There is layer upon layer here, let me think this through before I comment straight off the bat again :)
The step size is only zero here if we set it to zero in the else clause, so this is what the "recursive" below refers to, right? I think this works!

janzill: However, we would need to throw before setting the step size with the old scipy method on line 379 to be consistent, right?

Issue #128, opened 2020-04-17
Not possible to create graphs when no centroids are available in the model


pedrocamargo:
2020-04-17T11:37:50     WARNING    Traceback (most recent call last):
              File "C:/Users/pedro/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\AequilibraE-GUI\paths_procedures\show_shortest_path_dialog.py", line 92, in prepare_graph_and_network
              self.project.network.build_graphs()
              File "C:\Users\pedro\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\AequilibraE-GUI\aequilibrae\aequilibrae\project\network\network.py", line 257, in build_graphs
              g.prepare_graph(centroids)
              File "C:\Users\pedro\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\AequilibraE-GUI\aequilibrae\aequilibrae\paths\graph.py", line 226, in prepare_graph
              raise ValueError("Centroids need to be an array of integers 64 bits")
             ValueError: Centroids need to be an array of integers 64 bits

PR #131, opened 2020-04-17
Adds the possibility of having models with no centroids in the project file

pedrocamargo:
Adds the possibility of having models with no centroids in the project file
Refactoring of the graph class means that this can be used without the project as well

### DIFF
--- a/aequilibrae/project/network/network.py
+++ b/aequilibrae/project/network/network.py
@@ -276,7 +276,7 @@ class Network():
             warn(f'Fields were removed form Graph for being non-numeric: {",".join(removed_fields)}')
 
         curr.execute('select node_id from nodes where is_centroid=1;')
-        centroids = np.array([i[0] for i in curr.fetchall()])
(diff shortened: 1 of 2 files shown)


pedrocamargo:
Addresses issue #128

aequilibrae/project/network/network.py
```diff
@@ -276,7 +276,7 @@ def build_graphs(self) -> None:
             warn(f'Fields were removed form Graph for being non-numeric: {",".join(removed_fields)}')
 
         curr.execute('select node_id from nodes where is_centroid=1;')
-        centroids = np.array([i[0] for i in curr.fetchall()])
+        centroids = np.array([i[0] for i in curr.fetchall()], np.int32)
```
janzill, line 279: type here is different from graph.py, is that on purpose?

pedrocamargo: Not on purpose. Good catch

pedrocamargo:
Thanks for the review, @janzill !!

PR #132, opened 2020-04-27
Specific links exclusion

pedrocamargo:
Adds a new feature where one can "deactivate" links from a graph


pedrocamargo:
Thanks, @janzill !

Issue #116, opened 2020-03-19
Parameter write_back error


alexdawn:
I'm trying to get the example project to run in docker and conda to make use of jupyter notebooks, running the cell:

```python
import aequilibrae
from os.path import join
from aequilibrae import Parameters

p = Parameters()
p.parameters['system']['logging_directory'] =  fldr
p.write_back()
```

I'm getting the following error:

```
---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
<ipython-input-14-aff623fa8871> in <module>
      5 p = Parameters()
      6 p.parameters['system']['logging_directory'] =  fldr
----> 7 p.write_back()

/opt/conda/lib/python3.7/site-packages/aequilibrae/parameters.py in write_back(self)
     39     def write_back(self):
     40         """Writes the parameters back to file"""
---> 41         stream = open(self.file, "w")
     42         yaml.dump(self.parameters, stream, default_flow_style=False)
     43         stream.close()

PermissionError: [Errno 13] Permission denied: '/opt/conda/lib/python3.7/site-packages/aequilibrae/parameters.yml'
```

As far as I can tell the library is trying to rewrite itself which sounds a little dubious to me given Idempotency is always desirable. Is there a way of storing/reading the parameters elsewhere?

## Steps to reproduce

make the dockerfile:
```docker
FROM jupyter/scipy-notebook

USER root

RUN pip install aequilibrae
RUN pip install openmatrix
RUN apt-get update && apt-get -y upgrade
RUN apt-get install -y sqlite3 libsqlite3-dev libsqlite3-mod-spatialite
```

and run it with the bash command:

```bash
docker build --rm -t jupyter/aequilibrae .
docker run -p 8888:8888 -v "$PWD":/home/jovyan/work jupyter/aequilibrae start.sh jupyter notebook --NotebookApp.token=''
```

with the docker container now running jupyter navigate to http://localhost:8888 and try running the top snippet of python

pedrocamargo:
This is indeed an issue, @alexdawn . I don't see this as an idempotency issue. 
Right now AequilibraE is designed to write the parameters back to its origin. 
The problem with saving it in a temporary folder is that you may end up getting those parameters during a future session without noticing it.
Any suggestions on how to deal with that in a more robust way?

alexdawn:
[...]

PR #138, opened 2020-05-10
Re-structures project with some new features

pedrocamargo:


pedrocamargo:
@jamiecook  & @janzill , this one is HUGE, so take your time!!!

aequilibrae/parameters.py
```diff
@@ -31,21 +33,29 @@ class Parameters:
 
     def __init__(self):
         """ Loads parameters from file. The place is always the same. The root of the package"""
-        path = os.path.dirname(os.path.realpath(__file__))
-        self.file = os.path.join(path, "parameters.yml")
-        with open(self.file, "r") as yml:
-            self.parameters = yaml.load(yml, Loader=yaml.SafeLoader)
+
+        proj_path = os.environ.get('AEQUILIBRAE_PROJECT_PATH', '')
```
jamiecook, line 37: it's nice having an option to load from a path in an env var, but isn't there any other way to pass that information into this class... like arguments on the init method?

aequilibrae/paths/network_skimming.py
```diff
@@ -68,16 +68,16 @@ def execute(self):
             self.skimming.emit(["text skimming", "Saving Outputs"])
             self.skimming.emit(["finished_threaded_procedure", None])
 
-    def func_skim_thread(self, O, all_threads):
```
jamiecook, line 71: this seems like the same function as func_assign_thread, with exactly the same mistake (0 -> orig) in both.

Could be extracted and then future mistakes fixed in only one place :)

aequilibrae/paths/traffic_assignment.py
```diff
@@ -29,7 +29,7 @@ class TrafficAssignment(object):
         demand.computational_view(['matrix']) # We will only assign one user class stored as 'matrix' inside the OMX file
 
         project = Project()
```
jamiecook, line 31: I'd always be inclined to use construction to do your init/load ([RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization))

so setting up to use either
```python
project = Project(project_dir)
```

or 

```python
project = Project.from_dir(project_dir)
```

aequilibrae/project/network/database_triggers/link_type_table_triggers.sql
```diff
@@ -0,0 +1,58 @@
...
+-- Prevents a link_type record to be removed when it is in use for any link
+CREATE TRIGGER link_type_keep_if_in_use_deleting BEFORE DELETE ON "link_types"
+WHEN
+(Select count(*) from links where old.link_type = link_type)>0
+
+BEGIN
+    SELECT RAISE(ABORT, 'Link_type is in use on your network. Cannot change it');
+END;
+
+#
```
jamiecook, line 42: nice

aequilibrae/project/network/network.py
```diff
@@ -387,3 +288,8 @@ def add_spatial_index(self) -> None:
         curr.execute("""SELECT CreateSpatialIndex( 'links' , 'geometry' );""")
         curr.execute("""SELECT CreateSpatialIndex( 'nodes' , 'geometry' );""")
         self.conn.commit()
+
+    def __count_items(self, field: str, table: str, condition: str) -> int:
+        c = self.conn.cursor()
+        c.execute(f"""select count({field}) from {table} where {condition};""")
+        return c.fetchone()[0]
```
jamiecook, line 295: also nice

aequilibrae/project/project.py
```diff
@@ -28,58 +32,97 @@ def __init__(self):
...
+    def new(self, project_path: str) -> None:
         """Creates a new project
 
         Args:
-            *file_name* (:obj:`str`): Full path to the project data file. If file exists, it will fail
+            *project_path* (:obj:`str`): Full path to the project data folder. If folder exists, it will fail
         """
-        self.path_to_file = file_name
+        if self.__other_project_still_open():
+            raise Exception('You already have a project open. Close that project before creating a new one')
```
jamiecook, line 66: oof, why are you using global singleton type patterns?
What resource are you protecting that only allows one instance of project to be created at once?

aequilibrae/project/project.py
```diff
@@ -28,58 +32,97 @@ def __init__(self):
...
-
-        for mode in modes:
-            nm = list(mode.keys())[0]
-            descr = mode[nm]["description"]
-            mode_id = mode[nm]["letter"]
-            par = [f'"{p}"' for p in [nm, mode_id, descr]]
-            par = ",".join(par)
-            sql = f"INSERT INTO 'modes' (mode_name, mode_id, description) VALUES({par})"
-            cursor.execute(sql)
+        cursor.execute('PRAGMA foreign_keys = ON;')
```
jamiecook, line 121: did I show you the way we automatically setup connections to have this turned on?
```python
    engine = create_engine("sqlite:///{}".format(db_path))
    listen(engine, "connect", engine_listener)
```

where 

```python
def engine_listener(dbapi_conn, _connection_record):
    """Set up a listener to augment sqlalchemy connections with spatialite and foreign_key support."""
    add_mod_spatialite_path()
    dbapi_conn.enable_load_extension(True)
    dbapi_conn.load_extension("mod_spatialite")

    # Turn on foreign key checking
    cursor = dbapi_conn.cursor()

    cursor.execute("PRAGMA foreign_keys=ON")
    cursor.close()
```

aequilibrae/project/project_creation.py
```diff
@@ -0,0 +1,218 @@
...
+    create_modes_table(conn, parameters)
+    create_link_type_table(conn, parameters)
+    create_network_tables(conn, parameters)
+    populate_meta_extra_attributes(conn)
+    add_triggers(conn)
+
+
+def create_meta_table(conn) -> None:
+    cursor = conn.cursor()
+    create_query = f"""CREATE TABLE '{meta_table}' (name_table  VARCHAR NOT NULL,
```
jamiecook, line 22: have you considered sqlalchemy?

aequilibrae/project/project_creation.py
```diff
@@ -0,0 +1,218 @@
...
+    owlf = ["{} {}".format(fkey(f), f[fkey(f)]["type"]) for f in flds if fkey(f).lower() not in req_link_flds]
+
+    flds = fields["two-way"]
+    twlf = []
+    for f in flds:
+        nm = fkey(f)
+        tp = f[nm]["type"]
+        twlf.extend([f"{nm}_ab {tp}", f"{nm}_ba {tp}"])
+
+    link_fields = owlf + twlf
```
jamiecook, line 152: I feel that this is very opaque and could be commented / variable named better

aequilibrae/project/project_creation.py
```diff
@@ -0,0 +1,218 @@
...
+
+    if link_fields:
+        sql = sql.format(",".join(link_fields) + ",")
+    else:
+        sql = sql.format("")
```
jamiecook, line 157: I would make this

```suggestion
    if link_fields:
        link_field_definitions = ",".join(link_fields)
    else:
        link_field_definitions = ""
        
     sql = "big sql definition".format(link_field_definitions)
```

aequilibrae/project/project_creation.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # Run one query/command at a time
+    for cmd in query_list.split("#"):
+        try:
+            curr.execute(cmd)
+        except Exception as e:
+            msg = f"Error creating trigger: {e.args}"
+            logger.error(msg)
+            logger.info(cmd)
+    conn.commit()
```
jamiecook, line 218: so hacky - i love it

I've written the same method before... except I split on ;

tests/aequilibrae/project/test_network.py
```diff
@@ -58,43 +51,22 @@ def test_create_from_osm(self):
...
+        if items != 24:
+            self.fail('Wrong number of centroids found')
 
-        for f in flds:
-            if f not in nfields:
-                self.fail(f"Field {f} not added to nodes table")
+    def test_count_links(self):
+        items = self.siouxfalls.network.count_links()
+        if items != 76:
+            self.fail('Wrong number of links found')
```
jamiecook, line 67: isn't there a 
```
self.assert_equal(76, items, 'Wrong number of links found')
```

aequilibrae/parameters.py
pedrocamargo: The problem is that this is imported in different places by different procedures, and they would all need to get from the project that is being used.

This is also transient (it does not add anything permanently to the host system)

aequilibrae/paths/traffic_assignment.py
pedrocamargo: It used to be like that, but the API design around the creation of new projects and opening of existing ones turned out to be kinda weird. Any suggestions?

aequilibrae/project/network/network.py
pedrocamargo: Your suggestion from a while ago...

aequilibrae/project/project.py
pedrocamargo: This is motivated by being able to have the parameters file belonging to a project, instead of being global.

https://github.com/AequilibraE/aequilibrae/issues/116

aequilibrae/project/project.py
pedrocamargo: SQLAlchemy is not available with QGIS for Windows...

aequilibrae/project/project_creation.py
```diff
@@ -0,0 +1,218 @@
...
+    cursor.execute(create_query)
+    modes = parameters["network"]["modes"]
+
+    for mode in modes:
+        nm = list(mode.keys())[0]
+        descr = mode[nm]["description"]
+        mode_id = mode[nm]["letter"]
+        par = [f'"{p}"' for p in [nm, mode_id, descr]]
+        par = ",".join(par)
+        sql = f"INSERT INTO 'modes' (mode_name, mode_id, description) VALUES({par})"
```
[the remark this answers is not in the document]
pedrocamargo: I didn't know what an ORM was until 30s ago...   But I remember seeing a lot of that in the GHTS code...

aequilibrae/project/project_creation.py
pedrocamargo: I can't split on ';' because the triggers have ';' inside them....

tests/aequilibrae/project/test_network.py
pedrocamargo: There is...

aequilibrae/paths/network_skimming.py
pedrocamargo: The underlying function is actually quite different.

aequilibrae/project/project_creation.py
pedrocamargo: Not available inside QGIS for Windows

aequilibrae/project/project_creation.py
pedrocamargo: Improved a bit...

PR #149, opened 2020-05-13
About table

pedrocamargo:
API documentation coming in the following Pull request

### DIFF
--- a/aequilibrae/project/about.py
+++ b/aequilibrae/project/about.py
@@ -1,0 +1,74 @@
+import sqlite3
+from copy import deepcopy
+from aequilibrae.project.project_creation import create_about_table
+
+class About:
+    """Provides an interface for querying and editing the **about** table of an AequilibraE project"""
+    def __init__(self, conn: sqlite3.Connection):
+        self.__characteristics = []
+        self.__conn = conn
+        if self.__has_about():
+            self.__load()
+
+
+    def create(self):
+        """Creates the 'about' table for project files that did not previously contain it"""
+        if not self.__has_about():
+            create_about_table(self.__conn)
+            self.__load()
+
+    def add_info_field(self, info_field: str) -> None:
+        """Adds new information field to the model
+
+            Args:
+                *info_field* (:obj:`str`): Name of the desired information field to be added.  Has to be a valid
+                Python VARIABLE name (i.e. letter as first character, no spaces and no special characters)
+
+            ::
+
+                p = Project()
+                p.open('my/project/folder')
+                p.about.add_info_field('my_super_relevant_field')
+                p.about.my_super_relevant_field = 'super relevant information'
+                p.about.write_back()
+        """
+
+        if info_field[0].isalpha() and ' ' not in info_field:
+            sql = "INSERT INTO 'about' (infoname) VALUES(?)"
+            curr = self.__conn.cursor()
+            curr.execute(sql, info_field)
+            self.__conn.commit()
+        else:
+            raise ValueError(f'{info_field} is not valid as a metadata field.')
+
+    def write_back(self):
+        """Saves the information parameters back to the project database
+
+            ::
+
+                p = Project()
+                p.open('my/project/folder')
+                p.about.description = 'This is the example project. Do not use for forecast'
+                p.about.write_back()
+        """
+        curr = self.__conn.cursor()
+        for k in self.__characteristics:
+            v = self.__dict__[k]
+            curr.execute(f"UPDATE 'about' set infovalue = '{v}' where infoname='{k}'")
+        self.__conn.commit()
+
+    def __has_about(self):
+        curr = self.__conn.cursor()
+        curr.execute("SELECT name FROM sqlite_master WHERE type='table';")
+        if 'about' in [x[0] for x in curr.fetchall()]:
+            return True
+        return False
+
+    def __load(self):
+        self.__characteristics = []
+        curr = self.__conn.cursor()
+        curr.execute("select infoname, infovalue from 'about'")
+
+        for x in curr.fetchall():
+            self.__characteristics.append(x[0])
+            self.__dict__[x[0]] = x[1]
--- a/aequilibrae/project/about.py
+++ b/aequilibrae/project/about.py
@@ -2,14 +2,15 @@ import sqlite3
 from copy import deepcopy
 from aequilibrae.project.project_creation import create_about_table
 
+
 class About:
     """Provides an interface for querying and editing the **about** table of an AequilibraE project"""
+
     def __init__(self, conn: sqlite3.Connection):
         self.__characteristics = []
         self.__conn = conn
         if self.__has_about():
             self.__load()
-
 
     def create(self):
         """Creates the 'about' table for project files that did not previously contain it"""
--- a/aequilibrae/project/about.py
+++ b/aequilibrae/project/about.py
@@ -68,9 +68,7 @@ class About:
     def __has_about(self):
         curr = self.__conn.cursor()
         curr.execute("SELECT name FROM sqlite_master WHERE type='table';")
-        if 'about' in [x[0] for x in curr.fetchall()]:
-            return True
-        return False
+return any(['about' in x[0] for x in curr.fetchall()])
 
     def __load(self):
         self.__characteristics = []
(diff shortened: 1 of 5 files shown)


pedrocamargo:
@janzill

aequilibrae/project/about.py
```diff
@@ -0,0 +1,84 @@
...
+
+                p = Project()
+                p.open('my/project/folder')
+                p.about.add_info_field('my_super_relevant_field')
+                p.about.my_super_relevant_field = 'super relevant information'
+                p.about.write_back()
+        """
+        passed = True
+        for x in info_field:
+            if x not in string.ascii_lowercase + '_':
```
jamiecook, line 42: not a big deal here, but this will be constructing a new, merged, array each iteration, checking if x is in it and then throwing it away. Obviously this isn't high performance code so not crucial here, but just a habit to avoid.

jamiecook:
Reading through the About class... have you considered just using an ORM type helper like sqlalchemy?
I feel that most of that would be simply expressed as a "project has many comments"

aequilibrae/project/project_creation.py
```diff
@@ -121,15 +124,16 @@ def create_network_tables(conn, parameters) -> None:
...
+    lnk_tbl_sql = """CREATE TABLE 'links' (ogc_fid   INTEGER PRIMARY KEY,
+                                           link_id   INTEGER UNIQUE,
+                                           a_node    INTEGER,
+                                           b_node    INTEGER,
+                                           direction INTEGER NOT NULL DEFAULT 0,
+                                           distance  NUMERIC,
+                                           modes     TEXT    NOT NULL,
+                                           link_type TEXT REFERENCES link_types(link_type)
+                                           ON UPDATE RESTRICT ON DELETE RESTRICT {});
+                                           """
```
jamiecook, line 136: This isn't really "About" functionality

tests/data/no_triggers_project/parameters.yml
```diff
@@ -0,0 +1,208 @@
...
+        mode_filter:
+          pedestrian: 'no'
+        unknown_tags: true
+system:
+  cpus: 0
+  default_directory: C:\Users\pedro.camargo\Documents\sourcecode\drt
+  driving side: right
+  logging: true
+  logging_directory: D:/no_triggers_project
+  spatialite_path: C:\Users\pedro\Documents\mod_spatialite-NG-win-amd64
```
jamiecook, line 207: Do these hardcoded paths cause problems?

tests/aequilibrae/project/test_project_triggers.py
```diff
@@ -0,0 +1,191 @@
...
+    def test_link_type_triggers(self):
+        root = os.path.dirname(os.path.realpath(__file__)).replace('tests', '')
+        qry_file = os.path.join(root, "network/database_triggers/link_type_table_triggers.sql")
+        with open(qry_file, "r") as sql_file:
+            query_list = sql_file.read()
+            query_list = [cmd for cmd in query_list.split("#")]
+
+            def reboot_cursor():
+                self.proj.conn.commit()
+                self.curr = self.proj.conn.cursor()
```
jamiecook, line 33: this shouldn't be nested in the with block

tests/aequilibrae/project/test_project_triggers.py
```diff
@@ -0,0 +1,191 @@
...
+
+                self.curr.execute(cmd)
+                reboot_cursor()
+
+                with self.assertRaises(sqlite3.IntegrityError):
+                    sql = "UPDATE 'link_types' SET link_type_id= 'ww' where link_type_id='w'"
+                    self.curr.execute(sql)
+                reboot_cursor()
+
+            elif 'link_type_single_letter_insert' in cmd:
```
jamiecook, line 48: I'm not sure I like this setup, you seem to be running through a list of sql commands from file and based on the content of those, you are running some other sql first to set things up for the test. 

Why not just embed all the sql here in the test instead of having it in two places.

aequilibrae/project/about.py
pedrocamargo: > Reading through the About class... have you considered just using an ORM type helper like sqlalchemy?
> I feel that most of that would be simply expressed as a "project has many comments"

I did think about it, but QGIS for Windows does not ship with any ORM. This compatibility criterium is a real bitch, but it is one I need to keep for now.

tests/data/no_triggers_project/parameters.yml
pedrocamargo: They don't. There are workflows for them to be changed, and one of them will be deprecated with the upcoming version

tests/aequilibrae/project/test_project_triggers.py
pedrocamargo: Agreed.  I changed how it works for the modes and link_types tables yesterday. I will do the same for these

aequilibrae/project/project_creation.py
pedrocamargo: It slipped...

aequilibrae/project/about.py
pedrocamargo: > not a big deal here, but this will be constructing a new, merged, array each iteration, checking if x is in it and then throwing it away. Obviously this isn't high performance code so not crucial here, but just a habit to avoid.

Fair point.  It should look better now.

tests/aequilibrae/project/test_project_triggers.py
pedrocamargo: I refactored this completely already. It is in a different pull request

tests/aequilibrae/project/test_project_triggers.py
pedrocamargo: You can check it how I did it here:
https://github.com/AequilibraE/aequilibrae/pull/150/files?file-filters%5B%5D=.py&file-filters%5B%5D=.rst&file-filters%5B%5D=.sql&file-filters%5B%5D=.sqlite&file-filters%5B%5D=.txt#diff-bad12097672754322ad36a32502d7b54

pedrocamargo:
@jamiecook , it should look a little better now (the needed changes are in another pull request, actually).

aequilibrae/project/about.py
```diff
@@ -0,0 +1,82 @@
...
+                p = Project()
+                p.open('my/project/folder')
+                p.about.add_info_field('my_super_relevant_field')
+                p.about.my_super_relevant_field = 'super relevant information'
+                p.about.write_back()
+        """
+        allowed = string.ascii_lowercase + '_'
+        passed = [x for x in info_field if x not in allowed]
+
+        if not passed:
```
jamiecook, line 43: This looks the wrong way around... if not passed insert it?
Also, is passed actually a "fixed" version of info_field? passed implies a boolean to me

aequilibrae/project/about.py
```diff
@@ -0,0 +1,82 @@
...
+            v = self.__dict__[k]
+            curr.execute(f"UPDATE 'about' set infovalue = '{v}' where infoname='{k}'")
+        self.__conn.commit()
+
+    def __has_about(self):
+        curr = self.__conn.cursor()
+        curr.execute("SELECT name FROM sqlite_master WHERE type='table';")
+        if 'about' in [x[0] for x in curr.fetchall()]:
+            return True
+        return False
```
jamiecook, line 73:
```suggestion
return any(['about' in x[0] for x in curr.fetchall()])
```

aequilibrae/project/about.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __load(self):
+        self.__characteristics = []
+        curr = self.__conn.cursor()
+        curr.execute("select infoname, infovalue from 'about'")
+
+        for x in curr.fetchall():
```
jamiecook, line 80: I like to use col names in a pattern match here

```
for infoname,infovalue in curr.fetchall():
```

It avoids x[0] access which is always a headscratcher later on

aequilibrae/project/about.py
pedrocamargo: The problem here is that the variable for this list has the wrong name.  It should be called has_forbidden_characters. Should be clear now

aequilibrae/project/about.py
pedrocamargo: What do you mean?

PR #156, opened 2020-06-21
Link type table

pedrocamargo:
Adds the link type table to the project. Contains
 - Table creation
 - Consistency triggers
 - Table manipulation API


aequilibrae/project/network/link_type.py
```diff
@@ -0,0 +1,73 @@
...
+        conn = database_connection()
+        curr = conn.cursor()
+
+        curr.execute('pragma table_info(link_types)')
+        table_struct = curr.fetchall()
+        self.__fields = [x[1] for x in table_struct]
+        self.__original__ = {}
+
+        # data for the link_type
+        curr.execute(f"select * from 'link_types' where link_type_id='{link_type_id}'")
```
jamiecook, line 24: I'm a big fan of id as the id field, rather than "{table_name}_id"

jamiecook: I would also consider not having your __init__ hit the database

I like a pattern of __init__ just storing data into the relevant fields - at the moment - this class doesn't feel very "link_type" specific... most of it is very generic - the structure of the linkType table basically defines the fields in this class. I'd consider moving that logic into a generic databaseTableToClass field. 

Also the fact that you run a query per row in the database seems weird... I'd try to run a single query - select * from link_types - and then create a record for each and store all the records in a dict.

aequilibrae/project/network/link_type.py
```diff
@@ -0,0 +1,73 @@
...
+
+        conn = database_connection()
+        curr = conn.cursor()
+
+        curr.execute(f'select count(*) from link_types where link_type_id="{self.link_type_id}"')
+        if curr.fetchone()[0] == 0:
+            raise ValueError("Link type does not exist in the model. You need to explicitly add it")
+
+        curr.execute('pragma table_info(link_types)')
+        table_struct = [x[1] for x in curr.fetchall()]
```
jamiecook, line 64: probably should be a method

aequilibrae/project/network/link_types.py
```diff
@@ -0,0 +1,112 @@
...
+        self.curr.execute(f"select link_type_id from 'link_types' where link_type='{link_type}'")
+        found = self.curr.fetchone()
+        if len(found) == 0:
+            raise ValueError(f'Link type {link_type} does not exist in the model')
+        return LinkType(found[0])
+
+    def all_types(self) -> dict:
+        """Returns a dictionary with all mode objects available in the model. mode_id as key"""
+        self.__update_list_of_link_types()
+        return {x: LinkType(x) for x in self.__all_types}
```
jamiecook, line 108: This will trigger a lot of DB queries that could all be done in one hit

pedrocamargo: Do you see a way of doing it in one hit without duplicating the creation of LinkType objects?
I am also not too worried about this one, as its use will most likely be rare and the size of that table (link_types) should be incredibly small.

aequilibrae/project/network/link_type.py
pedrocamargo: The idea would be to create a generic class that would be subclassed for each table (whenever necessary)?

On the creation of items, are you suggesting that I generate all LinkType (and Mode, as they are very similar in structure) objects at once (say whenever the 1st one is requested by the user)?

aequilibrae/project/network/link_types.py
pedrocamargo: I think the idea is what I mention in the comment above?

aequilibrae/project/network/link_type.py
jamiecook: I would not sub-class it,  but yeah

```
class TableLoader():
    def load_table(curr, table_name):
        fields = get_table_struct(curr)
        curr.execute(f"select * from '{table_name}'")
        return [dict(zip(fields,row)) for row in curr.fetchall()]
```

Then you can just use that class in your LinkTypes and Modes classes. Prefer has_a to is_a in almost every circumstance.

pedrocamargo: @jamiecook  Take a look at what I did now. I took the core of your suggestions and got the LinkTypes and LinkType classes a lot cleaner and more robust, I Think.

aequilibrae/project/network/link_types.py
pedrocamargo: Should be solved

aequilibrae/project/network/link_type.py
pedrocamargo: Now there is no much reason why there couldn't be a generic class such as 

```
class GeneralObject:
    def __init__(self, data_set: dict) -> None:
        self.__original__ = {}
        for k, v in data_set.items():
            self.__dict__[k] = v
            self.__original__[k] = v
```

I would subclass that in both LinkType and Mode objects, but not sure if it makes enough sense to do it...

aequilibrae/project/database_specification/triggers/link_type_table_triggers.sql
```diff
@@ -23,27 +23,25 @@ END;
 
 CREATE TRIGGER link_type_keep_if_in_use_updating BEFORE UPDATE OF link_type ON "link_types"
 WHEN
-(Select count(*) from links where old.link_type = link_type)>0
-
+    (SELECT count(*) FROM links WHERE old.link_type = link_type)>0
 BEGIN
-    SELECT RAISE(ABORT, 'Link_type is in use on your network. Cannot change it');
+    SELECT RAISE(ABORT, 'Link_type is in use ON your network. Cannot change it');
```
jamiecook, line 28: weird capitalisation

aequilibrae/project/network/link_types.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError('Link Type ID already exists in the model. It must be unique.')
+
+        tp = {key: None for key in self.__properties}
+        tp['link_type_id'] = link_type_id
+        lt = LinkType(tp)
+        self.__items[link_type_id] = lt
+        logger.warning('Link type has not yet been saved to the database. Do so explicitly')
+        return lt
+
+    def drop(self, link_type_id: str) -> None:
```
jamiecook, line 90: this should probably be delete - drop is a term generally used to refer to deleting a table/view/db entity

aequilibrae/project/network/link_types.py
```diff
@@ -0,0 +1,129 @@
...
+        """Get a link_type from the network by its **link_type** (i.e. name)"""
+        for lt in self.__items.values():
+            if lt.link_type.lower() == link_type.lower():
+                return lt
+
+    def all_types(self) -> dict:
+        """Returns a dictionary with all LinkType objects available in the model. link_type_id as key"""
+        return self.__items
+
+    def save(self):
```
jamiecook, line 118: I would probably put the save functionality on this class, not the individual items.
Provide a set method on this class so that you can manipulate things in memory then publish them (like the network.publish in emme)

aequilibrae/project/network/link_types.py
```diff
@@ -0,0 +1,129 @@
...
+        # we can even keep editing and save it directly once we have added it to the project
+        new_type.lanes = 3
+        new_type.lane_capacity = 1100
+        new_type.save()
+
+    """
+    __items = {}
+
+    def __init__(self, net):
+        self.__all_types = []
```
jamiecook, line 62: so much caching, it can't be necessary everywhere

I'd try to avoid it as a design pattern except where you make it explicit... like the concept of a network that you load/publish.

pedrocamargo: The idea is similar to Emme's Load/publish, so the caching is necessary. Should I bring the save() up to the Network object only, then?

aequilibrae/project/network/link_types.py
pedrocamargo: Agreed

aequilibrae/project/network/link_types.py
pedrocamargo: See comment below

aequilibrae/project/network/link_types.py
jamiecook: What I'm saying though is that you are using class level variables to do it. I wasn't really clear with my language though, so apologies for that. 

Consider this example below - the two print(a) lines should produce the same output right? I haven't changed a in between calling them - but... that's not what happens.

I would advise that you have a single instance which manages the storage of this and then pass that around instead of shoving it into class variables and tracking whether they have been instantiated or not. 

```python
class A(object):
    my_var = []

    def __init__(self, bar):
        self.bar = bar
        self.my_var.append(bar)

    def __repr__(self):
        return f"{self.bar} -> {self.my_var}"


a = A(9)
print(a)    # 9 -> [9]
b = A(7)
c = A(8)
print(a)    # 9 -> [9, 7, 8]
```

pedrocamargo: That was intentional.  What I don't want is two sets of link types (or modes) floating around

PR #173, opened 2020-10-05
Improving trigger performance and ability to change triggers for older projects

pedrocamargo:


pedrocamargo:
This is really for you, @arobrien

aequilibrae/project/project_creation.py
```diff
@@ -64,10 +64,12 @@ def remove_triggers(conn: Connection) -> None:
             for qry in cmd.split("\n"):
                 if qry[:2] == '--':
                     continue
+                while '  ' in qry:
```
arobrien, line 67: This is more elegantly solved with regular expressions, so we can build in extra whitespace, and case insensitivity into the regex
```python
# you can cmpile the regex outside the for loop outside the for loop
# we can limit this to the start of a line only if appropriate by using carat '^' in the regex like r'^\s+create....
create_drop_regex = re.compile(r'create\s+trigger\s+(\w+)', flags=re.I)

# search and capture trigger name in one step
m = re.search(create_drop_regex, qry)
if m:
    try:
        curr.execute(f'drop trigger if exists {m.group(1).lower()}')
    ....
```

aequilibrae/project/database_specification/triggers/network_triggers.sql
```diff
@@ -190,13 +190,13 @@ create trigger updated_link_geometry after update of geometry on links
...
-    delete from nodes
-    where node_id not in (
-      select a_node
-      from links
-      union all
-      select b_node
-      from links);
+	Delete from Nodes
+    where node_id = old.a_node and
+           (select count(*) from Links where a_node = old.a_node or b_node = old.a_node) < 1;
```
arobrien, line 195: good optimisation and cleanup here.

arobrien (commented):
Looks good, but a couple of suggestions to clean up how the drops are processed,

aequilibrae/project/project_creation.py
arobrien: I re-read the documentation, it looks like the re mondule caches recent regexes, so you don't need to bother pre-compiling them.

pedrocamargo: Done and tested

PR #175, opened 2020-10-05
[WIP] Adds support for Python 3.9 [WIP]

pedrocamargo:


pedrocamargo:
GitHub Actions does not support Python 3.9 yet, so we will wait for support to arrive so we can go through with this

pedrocamargo:
Currently depending on Shapely's release. Such release can be followed here: https://github.com/Toblerity/Shapely/issues/944

Issue #176, opened 2020-10-05
Skims for PathComputation are not making sense


pedrocamargo:
**Purpose of use**
Retrieving skims after path computation

**Describe the bug**
Skim arrays are indexed with the internal IDs and are skipping the graph.nodes_to_index relation

**To Reproduce**

nm = 'D:/release/countries/Albania'

p = Project()
p.open(nm)

curr = p.conn.cursor()
curr.execute('Select node_id from Nodes WHERE modes like "%c%" limit 1000')
nodes = list(set([x[0] for x in curr.fetchall()]))

logger.info('Building Graphs')
p.network.build_graphs(['distance'], modes = 'c')

g = p.network.graphs['c'] # type: Graph

logger.info('Preparing Graph')
g.prepare_graph(np.array(nodes))
g.set_blocked_centroid_flows(False)
g.set_graph('distance')
g.set_skimming(['distance'])

res = PathResults()
res.prepare(g)

res.compute_path(nodes[0], nodes[-1])

# to find the milepost for the destination, you would do this
res.milepost[-1]

#but the code below should yield the same, but doesn't.
res.skims[res.destination]

# Instead, you need to do the following
res.skims[g.nodes_to_indices[res.destination]]

**Expected behavior**
res.skims[res.destination] 

should be equal to 

res.milepost[-1]

pedrocamargo:
Disconnected nodes should also have np.inf as a value, instead of its numerical HUGE equivalent

PR #178, opened 2020-10-05
Addresses issue #176

pedrocamargo:


aequilibrae/paths/results/path_results.py
```diff
@@ -91,13 +104,20 @@ def prepare(self, graph: Graph) -> None:
...
-        self.skims = np.zeros((self.nodes, self.num_skims), self.__float_type)
+        if self.num_skims:
+            self.skims = np.empty((graph.all_nodes[-1] + 1, self.num_skims), self.__float_type)
+            self.skims.fill(np.inf)
+            self._skimming_array = np.zeros((self.nodes, self.num_skims), self.__float_type)
+        else:
+            self._skimming_array = np.zeros((1, 2), self.__float_type)
+
         self.__graph_id__ = graph.__id__
+        self.__graph_sum = 2 * graph.cost.sum()
```
jamiecook, line 120: why 2*?

aequilibrae/paths/results/path_results.py
```diff
@@ -107,7 +127,9 @@ def reset(self) -> None:
         if self.predecessors is not None:
             self.predecessors.fill(-1)
             self.connectors.fill(-1)
-            self.skims.fill(np.inf)
+            if self.skims is not None:
+                self.skims.fill(np.inf)
+            self._skimming_array.fill(np.inf)
```
jamiecook, line 132: should this be under the condition too?

jamiecook (approved):

aequilibrae/paths/results/path_results.py
pedrocamargo: Because I can imagine somebody creating a worst-case scenario network (imagining that turn penalties are considered) where one needs to traverse all links (or almost all) in both directions.

jamiecook: maybe worth commenting that?

aequilibrae/paths/results/path_results.py
pedrocamargo: No... they are always used to instantiate some code in Cython..

PR #179, opened 2020-10-10
Limits the number of CPU usage for the BPR according to user parameters

pedrocamargo:


pedrocamargo:
An easy one, @janzill

aequilibrae/paths/bpr.pyx
```diff
@@ -1,27 +1,31 @@
 from libc.math cimport pow
 from cython.parallel import prange
 
-def bpr(congested_times, link_flows, capacity, fftime, alpha, beta):
+def bpr(congested_times, link_flows, capacity, fftime, alpha, beta, cores):
+    cdef int c
```
janzill, line 5: Is there a reason you declare and initialise in different places?

janzill:
True, but I still have a question :D

aequilibrae/paths/bpr.pyx
pedrocamargo: Yes.  My poor programming skills. Solved now (and incorporated elsewhere)

janzill (approved):

janzill:
merge when passing tests

PR #181, opened 2020-10-10
Adds API to handle the log file

pedrocamargo:


aequilibrae/log/log.py
```diff
@@ -0,0 +1,38 @@
...
+        Return:
+            *log_contents* (:obj:`list`): List with all entries in the log file
+        """
+
+        with open(self.log_file_path, 'r') as file:
+            return [x.strip() for x in file.readlines()]
+
+    def clear(self):
+        """Clears the log file. Use it wisely"""
+        _ = open(self.log_file_path, 'w')
```
jamiecook, line 38: wont this keep an open handle around?

what about

```
with open() as _:
    pass
```

which should explicitly close it

jamiecook (approved):

aequilibrae/log/log.py
pedrocamargo: Good catch!

PR #185, opened 2020-10-25
Introduces comprehensive result treatment and saving for traffic assignment

pedrocamargo:


pedrocamargo:
This is a really cool one, @jamiecook

aequilibrae/paths/traffic_assignment.py
```diff
@@ -306,3 +329,126 @@ def __validate_parameters(self, kwargs) -> bool:
...
+            *info* (:obj:`dict`): Pandas dataframe with all the assignment results indexed on link_id
+        """
+
+        classes = {}
+        for cls in self.classes:
+            if len(cls.matrix.view_names) == 1:
+                classes[cls.graph.mode] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for nm in cls.matrix.view_names}
+            else:
+                classes[cls.graph.mode] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for i, nm in
+                                           enumerate(cls.matrix.view_names)}
```
jamiecook, line 436:
```suggestion
        def f(cls):
            return {nm: np.sum(cls.matrix.matrix_view[:, :]) for i,nm in enumerate(cls.matrix.view_names)}
        classes = {cls.graph.mode: f(cls) for cls in self.classes}            
```

jamiecook: I was just refactoring to show off how I've been using nested functions and dict comprehensions.... but then I couldn't figure out why you even have an if statement or why you are using enumerate

jamiecook (approved):

aequilibrae/paths/traffic_assignment.py
```diff
@@ -65,15 +74,21 @@ class TrafficAssignment(object):
         convergence_report = pd.DataFrame(assig.assignment.convergence_report)
         convergence_report.head()
 
-        # Link flow results are here
-        results = assigclass.results
+        # Assignment results can be viewed as a Pandas DataFrame
+        results_df = assig.results()
+
+        # information on the assignment setup can be recovered wuth
```
janzill, line 80: wuth -> woof. Or with :)

aequilibrae/paths/traffic_assignment.py
pedrocamargo: Because of that.  The array is two dimensional if only one user class per traffic class, and three dimensional when there are multiple user classes per traffic class

janzill: Are you forgetting to sum over the third dimension?

janzill (approved):

aequilibrae/paths/traffic_assignment.py
```diff
@@ -179,16 +198,20 @@ def set_algorithm(self, algorithm: str):
...
         if algorithm.lower() == "all-or-nothing":
             self.assignment = allOrNothing(self)
+        elif algorithm.lower() == 'fw':
+            self.assignment = LinearApproximation(self, "frank-wolfe")
         elif algorithm.lower() in ["msa", "frank-wolfe", "cfw", "bfw"]:
             self.assignment = LinearApproximation(self, algorithm.lower())
         else:
             raise Exception('Algorithm not listed in the case selection')
 
+        self.__dict__['algorithm'] = algorithm
```
janzill, line 213: I think this is why your test is failing, maybe set algorithm to 'frank-wolfe' if the user provided fw?

janzill: On the other hand special treatment is a bit annoying (as proven by the failing test :) ). Maybe we should wrap this up into its own class?

janzill: Not too important right now, but if we ever add other assignment methods we might have to

aequilibrae/paths/traffic_assignment.py
jamiecook: as in 
```
classes[cls.graph.mode] = { nm: np.sum(cls.matrix.matrix_view[:, :, i]) 
                         for i, nm in enumerate(cls.matrix.view_names)
                       }

aequilibrae/paths/traffic_assignment.py
jamiecook: special treatment is fine... as long as you just do it once.
do what Jan said, if algo == "fw" then replace algo = "frank-wolfe"

After that point never allow fw, only frank-wolfe

aequilibrae/paths/traffic_assignment.py
pedrocamargo: each dimension is a different slice that gets assigned within a single class, so we can assign multiple demand slices per traffic class

pedrocamargo: Your suggestion doesn't work because of this variation in the number of dimensions of the array, @jamiecook

aequilibrae/paths/traffic_assignment.py
pedrocamargo: I think it is solved appropriately now.

pedrocamargo:
Thanks, @janzill & @jamiecook !!

aequilibrae/paths/traffic_assignment.py
jamiecook: suggestion (anon method and dict comprehension) would still work... just needs your bug fix and the if statement

jamiecook: ... also... why not just always have a 3 dim matrix?
i.e. `matrix = np.ndarray(x,x,1)`

pedrocamargo: I don't remember now...  There was a reason when I implemented it...

PR #187, opened 2020-11-10
Assignment classes

pedrocamargo:


pedrocamargo:
This is a really small improvement, but it seemed a little too obvious not to make...

aequilibrae/paths/traffic_assignment.py
```diff
@@ -285,7 +303,7 @@ def set_cores(self, cores: int) -> None:
             cores (:obj:`int`): Number of CPU cores to use
         """
         self.cores = cores
-        if self.classes is not None:
+        if self.classes:
```
jamiecook, line 306: if classes is an array a for loop over an empty array does nothing so you don't really need to wrap it in an if statement

jamiecook (approved):

aequilibrae/paths/traffic_assignment.py
pedrocamargo: Setting the number of cores should be inside the if, actually...

jamiecook: ah... another nice trick is to not use `else raise`

jamiecook:
```
        if not self.classes:
            raise Exception('You need load traffic classes before overwriting the number of cores')
      
        # continue as per normal
        self.cores = cores
        for c in self.classes:
            c.results.set_cores(cores)
                c._aon_results.set_cores(cores)
```

pedrocamargo: Hahahahaha  I thought of the same thing while the tests ran locally... hahahaha

Issue #136, opened 2020-05-10
Adding Matrix table to the project


pedrocamargo:
Aequilibrae is missing a table with records and metadata for all matrices associated with a given project

PR #188, opened 2020-11-10
Adds matrix indexing to the project - Issue #136

pedrocamargo:
Built documentation is here:

http://aequilibrae.com.s3-website-us-east-1.amazonaws.com/python/dev/_generated/aequilibrae.project.Matrices.html#aequilibrae.project.Matrices


aequilibrae/project/data/matrix_record.py
```diff
@@ -0,0 +1,84 @@
...
+        conn.close()
+        del self.__items[self.name]
+        del self
+
+    def update_cores(self):
+        """Updates this matrix record with the matrix core count in disk"""
+        self.__dict__['cores'] = self.__get_cores()
+
+    def get_data(self) -> AequilibraeMatrix:
+        """Returns the actual matrix fort computation
```
jamiecook, line 55:
```suggestion
        """Returns the actual matrix for further computation
```

aequilibrae/project/data/matrices.py
```diff
@@ -0,0 +1,167 @@
...
+        tp = {key: None for key in self.__fields}
+        tp['name'] = name
+        tp['file_name'] = file_name
+        mat = AequilibraeMatrix()
+        mat.load(join(self.fldr, file_name))
+        tp['cores'] = mat.cores
+        mat.close()
+        del mat
+
+        mr = MatrixRecord(tp, self.__items, self.__project.logger)
```
jamiecook, line 159: are you storing a name cache in the Matrices class and then passing down a reference to that to each instance of the Record class as well.

I feel that's not a good idea... you'd be better off enforcing that constraint at either the database level or in your Matrices class

jamiecook:
Overall I like the structure, a few examples of some hard to understand variable names but the overall user experience of it should be pretty good.

aequilibrae/project/data/matrices.py
pedrocamargo: I had that feeling....

pedrocamargo: Just addressed

PR #193, opened 2020-11-24
Improves Network Skimming class with results saving to project database

pedrocamargo:


pedrocamargo:
This time I had another PR in line, @jamiecook . Though luck!!

pedrocamargo:
Since you are working a bit, @janzill !  :-D

tests/aequilibrae/paths/test_network_skimming.py
```diff
@@ -1,40 +1,75 @@
...
+        _ = skimming_single_origin(12, graph, res, aux_res, 0)
 
-        skm = NetworkSkimming(g, res)
+        skm = NetworkSkimming(graph)
         skm.execute()
 
-        tot = np.nanmax(res.skims.distance[:, :])
-
+        tot = np.nanmax(skm.results.skims.distance[:, :])
         if tot > 10e10:
```
felasouza, line 57: not sure the standard but I would have a constant defined with this number

aequilibrae/project/data/matrices.py
```diff
@@ -77,7 +77,7 @@ def update_database(self) -> None:
                 while f'{name}_{i}' in self.__items:
                     i += 1
                 name = f'{name}_{i}'
```
felasouza, line 79: is it saving matrices with increasing number, right? aaa_0, aaa_1. If so, it perhaps helps the user the zfill on the numbers so that it stays sorted. If this doesn't make sense, please ignore it :-)

felasouza (approved):

tests/aequilibrae/paths/test_network_skimming.py
pedrocamargo: This is largely arbitrary, as paths not found return something closer to infinite.  I have changed to something a little more sensible

aequilibrae/project/data/matrices.py
pedrocamargo: There is no need for matrices to be in order.  This  is just a way to generate an arbitrary matrix name when all the user does is throw the matrix in the appropriate folder

PR #194, opened 2020-11-24
Assignment skims can be saved directly to the project

pedrocamargo:


aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -80,13 +78,44 @@ def __init__(self):
...
         self.omx_file = None  # type: omx.File
         self.__version__ = VERSION  # Writes file version
 
+    def save(self, names=()) -> None:
+        """Saves matrix data back to file.
+
+        If working with AEM file, it flushes data to disk.  If working with OMX, requires new names
+
+        Args:
+            *names* (:obj:`List[str]`, `Optional`): New names for the matrices. Required if working with OMX files"""
```
pedrocamargo, line 94: The type of argument is a tuple, not list

pedrocamargo:
Can you review this one, @jamiecook ?

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -80,13 +78,44 @@ def __init__(self):
...
+            self.matrices.flush()
+            return
+
+        if isinstance(names, str):
+            names = [names]
+
+        if len(names) != len(self.view_names):
+            raise ValueError('Number of names needs to be equal to computational view')
+
+        exists = [n for n in names if n in self.names]
```
jamiecook, line 106: I sometimes like any for this

```suggestion
        exists = any([n in self.names for n in names]
```

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
                 'Target RGap': self.assignment.rgap_target}
         return info
 
-    def save_skims(self, matrix_name: str, which_ones='final') -> None:
+    def save_skims(self, name: str, which_ones='final', format='omx') -> None:
         """Saves the skims (if any) to the skim folder and registers in the matrix list
 
         Args:
-            matrix_name (:obj:`str`): Name of the file to hold this matrix
+            name (:obj:`str`): Name of the file to hold this matrix
```
jamiecook, line 487: i'd call this filename

pedrocamargo: fair

jamiecook: oh, except it's not until you add an "omx" / "aem" extension...

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
...
+        avg_skims = self.classes[0].results.skims  # type: AequilibraeMatrix
+
+        # The ones for the last iteration are here
+        last_skims = self.classes[0]._aon_results.skims  # type: AequilibraeMatrix
+
+        names = []
+        if which_ones in ['final', 'all']:
+            if last_skims.names:
+                for core in last_skims.names:
+                    names.append(f'{core}_final')
```
jamiecook, line 519:
```suggestion
        if which_ones in ['final', 'all'] and last_skims.names:
            for core in last_skims.names:
                names.append(f'{core}_final')
```

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
...
+
+        if which_ones in ['blended', 'all']:
+            if avg_skims.names:
+                for core in avg_skims.names:
+                    names.append(f'{core}_blended')
```
jamiecook, line 524:
```suggestion
        if which_ones in ['blended', 'all'] and avg_skims.names:
            for core in avg_skims.names:
                names.append(f'{core}_blended')
```

aequilibrae/paths/traffic_assignment.py
pedrocamargo: last_skims might be None...

Maybe I should change to an empty matrix?

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
...
+
+        if not names:
+            raise ValueError('No skims to save')
+        # Assembling a single final skim file can be done like this
+        # We will want only the time for the last iteration and the distance averaged out for all iterations
+        if mat_format == 'aem':
+            working_name = export_name
+        else:
+            working_name = AequilibraeMatrix().random_name()
```
jamiecook, line 533:
```suggestion
        working_name = export_name if mat_format == 'aem' else AequilibraeMatrix().random_name()
```

jamiecook: sometimes i think the ternary operator in python makes sense

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
...
+                  'matrix_names': names}
+
+        # Create the matrix to manipulate
+        out_skims = AequilibraeMatrix()
+        out_skims.create_empty(**kwargs)
+
+        out_skims.index[:] = self.classes[0].graph.centroids[:]
+        out_skims.description = f'Assignment skim from procedure ID {self.procedure_id}'
+
+        if which_ones in ['final', 'all']:
```
jamiecook, line 546: and

aequilibrae/paths/traffic_assignment.py
```diff
@@ -474,12 +480,89 @@ def info(self) -> dict:
...
+            if last_skims.names:
+                for core in last_skims.names:
+                    out_skims.matrix[f'{core}_final'][:, :] = last_skims.matrix[core][:, :]
+
+        if which_ones in ['blended', 'all']:
+            if avg_skims.names:
```
jamiecook, line 552: and

aequilibrae/project/data/matrices.py
```diff
@@ -1,22 +1,22 @@
...
     __fields = []
 
-    def __init__(self, project):
-        self.__project = project
-        self.conn = project.conn  # type: Connection
+    def __init__(self):
+        self.conn = database_connection()
         self.curr = self.conn.cursor()
-        self.fldr = os.path.join(project.project_base_path, 'matrices')
+        self.fldr = os.path.join(os.environ.get(environ_var), 'matrices')
```
jamiecook, line 19: where does environ_var come from?

jamiecook: oh it's imported? - is it a CONSTANT?

jamiecook (approved):

aequilibrae/project/data/matrices.py
pedrocamargo: Yes, it is

jamiecook: doesn't pep8 say that it should be ENVIRON_VAR?

aequilibrae/paths/traffic_assignment.py
jamiecook: not sure what you mean... last_skims will only be populated if which_ones == final or all?

if so that is what short circuits are for

jamiecook: image
```
>>> def foo():
...     raise ValueError("Dont call me")
...
>>>
>>> a = 7 and foo()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in foo
ValueError: Dont call me
>>> a = False and foo()
>>>
```

jamiecook: if the first arg to and is false it never evaluates the second
same with OR except the other way round (if first arg is true, no need to eval the second)

aequilibrae/project/data/matrices.py
pedrocamargo: I don't think so (just searched and could not find anything on it)

jamiecook: why oh why do you doubt me?
image

aequilibrae/paths/traffic_assignment.py
pedrocamargo: I guarantee that the skim is an AequilibraE matrix object, so the second if is no longer needed. This is probably the cleanest solution

aequilibrae/paths/traffic_assignment.py
pedrocamargo: same as above

aequilibrae/project/data/matrices.py
pedrocamargo: It has to do with being a constant!!! I was looking for references to environment variables!!!

And I doubt you just to keep you sharp! Fixed

aequilibrae/paths/traffic_assignment.py
pedrocamargo: I fixed in a more comprehensive way, where the skims object is always an instance of an AequilibraE matrix and it might just be an empty one

aequilibrae/paths/traffic_assignment.py
pedrocamargo: Same change as below applies

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: The list is used in the message inside the if...

PR #196, opened 2020-11-29
Refactors distribution

pedrocamargo:


pedrocamargo:
Do you want to take a stab at reviewing this one, @felasouza ?

aequilibrae/distribution/gravity_application.py
```diff
@@ -142,20 +144,25 @@ def __init__(self, **kwargs):
...
         self.core_name = kwargs.get("output_core", "gravity")
-        self.output_name = kwargs.get("output", AequilibraeMatrix().random_name())
+        self.output_name = AequilibraeMatrix().random_name()
         self.nan_as_zero = kwargs.get("nan_as_zero", False)
         self.output = None
         self.gap = np.inf
         self.logger = logging.getLogger("aequilibrae")
+        self.procedure_date = ''
+        self.procedure_id = ''
+        self.__ipf = None  # type: Ipf
```
felasouza, line 156: perhaps newbie question here. The choice of using double underscore may cause issue (or help to not have issues) when inhereting the class. I am not sure what would be the case here.

aequilibrae/distribution/gravity_application.py
```diff
@@ -188,17 +195,17 @@ def apply(self):
...
-        self.output = ipf.output
-        self.gap = ipf.gap
+        # apply Fratar
+        self.__ipf.fit()
+        self.output = self.__ipf.output
+        self.gap = self.__ipf.gap
 
-        q = ipf.report.pop(0)
-        for q in ipf.report:
+        _ = self.__ipf.report.pop(0)
```
felasouza, line 207: what does it do? what is there that must be removed?

janzill (approved):
Looks good to me

aequilibrae/distribution/ipf.py
```diff
@@ -237,6 +247,39 @@ def fit(self):
             self.report.append("")
             self.report.append("Running time: " + str("{:4,.3f}".format(perf_counter() - t)) + "s")
 
+    def save_to_project(self, name: str, file_name: str) -> MatrixRecord:
+        """Saves the matrix output to the project file
+
+        Args:
```
felasouza, line 253: out of curiosity and not sure if would cause more headache than benefit. But I've seen in the skim saving similar opeartions of checking stuff, checking existence and so on. Although straighforward it might be the case to handle this type of operations in a different class (which could check extension, existence and return the record or part of it)

felasouza (approved):

aequilibrae/distribution/gravity_application.py
pedrocamargo: Not an issue

aequilibrae/distribution/ipf.py
pedrocamargo: True

aequilibrae/distribution/gravity_application.py
pedrocamargo: Appends IPF report to the distribution report. Very old code, so it is just a bit dump.  Just replaced it

aequilibrae/distribution/ipf.py
pedrocamargo: How does it look like now?

felasouza: I liked.

PR #197, opened 2020-12-06
[DO NOT MERGE] Felipes bush based assignment

janzill:
@felasouza I have included your code pretty much one-to-one in paths/path_based. If you go to that directory, run
```bash
python setup_path_based.py build_ext --inplace
```
I haven't done much on integration with aequilibrae yet, I started setting things up in paths/path_based_assignment.py, but it's far from functional. This means path building is still done with boost, and problem loading with the implementation in OpenBenchmark.
@pedrocamargo , on ubuntu, install boost with 
```bash
sudo apt-get install libboost-all-dev
```
My goal is to replace boost with aequilibrae's Dijkstra implementation as discussed, so I won't integrate boost in the setup.

I've been working in two notebooks (in jupyter_development):

 1) path_based_integration.ipynb contains the code that used to be in futuretest/cythoncodetest.py. This is the starting point for my attempt to port things over to aequilibrae. I set up the two test cases you mentioned @felasouza and a comparison to aequilibrae's bi-conjugate Frank-Wolfe method. Unfortunately we cannot use the best known solution from the tntp repositiory because it it doesn't allow paths through centroids, which is functionality that is missing in this implementation. Luckily aequilibrae can do either, and when allowing flows through centroids the Anaheim solution aligns well. However, this is not true for Berlin-Center. As discussed, my goal was to integrate the algorithm in aequilibrae to make sure demand and networks are identical, as well as being able to block flows through centroids. This isn't too hard, but I got distracted by trying to implement the parallel algorithm, see 2).

 2) path_based_parallel.ipynb, with corresponding changes in the c++ code. This is my attempt at implementing the parallel algorithm. Something is not right here just yet, the objective value at the calculated solution is too large, I think it converges but to a different solution so there's an error somewhere. I've added stepsize finding, the derivative of the objective function w.r.t to the stepsize is implemented in c++ and the root finding is taken from scipy. @felasouza if you could have a look at the code and see if you can spot any mistakes that would be great. I have not parallelised anything, the notebook contains loops over origins, that bit is done best when integrating with aequilibrae.

I'm going to focus on getting the non-parallel version integrated with aequilibrae, probably by wrapping the data structures with a thin layer such that there's less work re-writing data structures in the path-based implementation for now. This can be done once we are satisfied that the algorithm is working, and hopefully together with the parallel implementation. Let me know what you think.


pedrocamargo:
This is VERY exciting!!  IU will work on this after I am done with a few things I need to get done today!!

felasouza:
Sounds great! This weekend I worked on a few things (mainly switching to
std::vector some of the raw pointer I had so that it consumes less memory).
Also working now in a potential improvement for larger instances...

Em dom., 6 de dez. de 2020 às 20:42, Pedro Camargo <notifications@github.com>
escreveu:

janzill:
I've wrapped the aequilibrae demand and network in a quick and dirty way such that we can now run both the (non-parallel) bush-based assignment as well as the linear approximation, see jupyter_development/integrate_path_based_in_aequ.ipynb

For Anaheim, this yields consistent results, but not for Berlin-Center. I don't know what's going on there, but with the current implementation and the notebook in jupyter_development/parsing_networks.ipynb (which I used to parse the demand into an aequilibrae matrix), we should be able to try other examples.

janzill:
I think the next step is to use Aequilibrae's shortest path algorithm instead of boost, quick and dirty in a first step to get it running quickly I'd say

pedrocamargo:
@janzill , I was looking into the notebooks you worked on, but could not find the BFW Berlin assignment. Looking into the network I noticed that the BPR parameters vary between two sets of values (0, 4.0) and (2.0, 4.0) for centroid connectors and otherwise, respectively.  These numbers aren't only weird, but they are also very different than the default (0.15, 4.0), which would result in massive differences in the assignment results.  Are those parameters the same for both assignments?

janzill:
@pedrocamargo I noticed this too, I made sure to use these for both methods. I also ran a test with all values set to 0.15 for alpha, which still lead to inconsistent results.

janzill:
image

janzill:
from https://github.com/AequilibraE/aequilibrae/blob/690339dd72db227f9dd0da484efbbf431d6de911/jupyter_development/integrate_path_based_in_aequ.ipynb

janzill:
Alright, I think I have eliminated boost as a dependency by integrating Aequilibrae's shortest path. And when I say integrating, I mean hacking. I wanted to get this done quickly and the data structures will have to change anyways, so I've wrapped it and duplicated calculations in places. It looks like we can block flows through centroids now, which means we can compare to the best known solution.

janzill:
See https://github.com/AequilibraE/aequilibrae/blob/1eaea27e10e62331ac24f9beeaa89fb482279a61/jupyter_development/integrate_shortest_path.ipynb for an example. It's a bit of a mess because it's a dev notebook, but I think we should set up numerical experiments and comparisons to make sure the method is working.

janzill:
Yes, blocking flows through centroids is working, this is what I get for Anaheim when running bush-based to 1e-6 and BFW to 5*1e-5. The plot compares the best known solution to bush-based. The difference columns in the table are for BFW compared to bush-based, they are sorted by largest absolute difference.

image

janzill:
ok, berlin center with flows blocked through centroids, BFW rgap of 1e-5, path gap of 1e-6 (see https://github.com/AequilibraE/aequilibrae/blob/c5280b21ac427ea36c7da0b34b73be16c70009e9/jupyter_development/compare_path_based.ipynb):
image

janzill:
@pedrocamargo I had a bit of time waiting for some calculations to finish and decided to remove (comment out for now) any boost dependency. This means it should now compile on any platform with just the requirements.txt. @felasouza I also copied in the bits on the parallel algorithm, something is still wrong, it does not converge to the same solution as the sequential algorithm

pedrocamargo:
@janzill , I merged the current dev version into this one in order to get the CI/CD going and try to get it into the 0.7 version.  Let's see what test results will bring

janzill:
I started having a look at integrating the data structures. Keeping it with the current structure as much as possible is obviously desirable, but if we want to avoid copying results between the numpy and the C++ path finder instance we need to make sure things are set up correctly. I don't think there's a problem with setting up most of the arrays as c-contiguous, but for threading there is some slicing happening and that makes things harder so I need to think about this a bit more when I have some time. There's also the possibility of bringing this all into cython, but I'm not sure if that would incur a runtime penalty. I haven't pushed anything because I want to keep this branch in a state where the assignment is working, I might start working on a different branch once I've given this a bit more thought.

pedrocamargo:
I know @felasouza did some advancement on this on Bitbucket as well, so it might be a good time to consolidate everything

felasouza (approved):
looks all fine. Just one small question is that in your initial message I though there would be a path_based folder (which you copied almost one to one my old code) within aequilibrae/paths. Am I looking for it in the wrong place?

janzill:
@felasouza When @pedrocamargo merged the dev branch for version 0.7 I made the file structure consistent with what was there before and got rid of that folder and all the files that are not used anymore.

aequilibrae/paths/link.py
```diff
@@ -0,0 +1,56 @@
...
+        self.link_index = None
+        self.t0 = t0
+        self.capacity = capacity
+        self.alfa = alfa
+        self.beta = beta
+        self.direction = direction
+        self.t0_ba = t0_ba
+        self.capacity_ba = c_ba
+
+    # import math
```
pedrocamargo, line 22: Shouldn't we delete all this commented stuff?

aequilibrae/paths/path_based_assignment.py
```diff
@@ -0,0 +1,441 @@
...
+    from aequilibrae.paths import TrafficAssignmentCy
+
+    # temp for one to all shortest path hack
+    from aequilibrae.paths.multi_threaded_aon import MultiThreadedAoN
+    from aequilibrae.paths.AoN import one_to_all
+except ImportError as ie:
+    logger.warning(f"Could not import procedures from the binary. {ie.args}")
+
+
+import scipy
```
pedrocamargo, line 26: Imports top the top

aequilibrae/paths/path_based_assignment.py
```diff
@@ -0,0 +1,441 @@
...
+    recent_scipy = False
+    logger.warning("Using older version of Scipy. For better performance, use Scipy >= 1.4")
+
+
+if False:
+    from aequilibrae.paths.traffic_assignment import TrafficAssignment
+
+spec = iutil.find_spec("PyQt5")
+pyqt = spec is not None
+if pyqt:
```
pedrocamargo, line 44: What if pyqt is not available?  Shouldn't we be importing the empty class?

aequilibrae/paths/link.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Link:
+    def __init__(self, node_id_from, node_id_to, link_id, t0, capacity, alfa, beta, direction=1, t0_ba=None, c_ba=None):
+        self.link_id = link_id
+        self.node_id_from = node_id_from
+        self.node_id_to = node_id_to
+        self.link_index = None
+        self.t0 = t0
+        self.capacity = capacity
+        self.alfa = alfa
```
pedrocamargo, line 16: We have other VDF coming (see https://github.com/AequilibraE/aequilibrae/issues/204) , so the structure needs to be a little more generic

aequilibrae/paths/path_based_assignment.py
```diff
@@ -0,0 +1,441 @@
...
+    from PyQt5.QtCore import pyqtSignal as SIGNAL
+
+
+class PathBasedAssignment(WorkerThread):
```
pedrocamargo, line 48: We could use some docstrings here, I guess

aequilibrae/paths/traffic_assignment.py
```diff
@@ -93,8 +94,9 @@ class TrafficAssignment(object):
         avg_skims = assigclass.results.skims # blended ones
         last_skims = assigclass._aon_results.skims # those for the last iteration
     """
+
     bpr_parameters = ["alpha", "beta"]
-    all_algorithms = ["all-or-nothing", "msa", "frank-wolfe", "fw", "cfw", "bfw"]
+    all_algorithms = ["all-or-nothing", "msa", "frank-wolfe", "fw", "cfw", "bfw", "bush-based", "bush-based-parallel"]
```
pedrocamargo, line 99: Is the parallel version ready and coming with this PR?

jupyter_development/utils.py
```diff
@@ -0,0 +1,91 @@
+import os
```
pedrocamargo, line 1: Is this script plus all the notebooks to be merged?  Aren't they just dev stuff?

pedrocamargo:
Things look pretty good, but I guess we still need documentation and cleaning of the PR, as there seems to be a lot of dev scripts/notebooks mixed in with the actual library/code

janzill:
@pedrocamargo there is quite a bit of work to do here I think, the most important thing being getting rid of duplicate calculations. Off the top of my head, there are two major things that need to change:

  1) The BPR VDF is hard-coded in the C++ parts of the code, this will have to be removed so that we only make use of the current vdf class.

  2) Improve the handling of path precedence array

  3) Probably get rid of the link class and use the underlying aequilibrae graph numpy arrays

Both of these are probably best handled by making the underlying numpy arrays c-contiguous and then using these directly in the C++ parts, see one of the many (many) comments above. However, I haven't given it any further thought.

If the goal is to get this working as quickly as possible we could go with what we have, but I'd rather spend a bit more time on this and make it consistent with what's there already, otherwise it's going to be a nightmare to maintain and possibly extend, the vdf issue you mentioned @pedrocamargo being a prime example

aequilibrae/paths/link.py
janzill: Agreed, in fact I'd like to remove the link class altogether

aequilibrae/paths/traffic_assignment.py
janzill: Good point, it's not ready and should be removed

jupyter_development/utils.py
janzill: they are, I thought about making them gists but then went with having everything in one place for collaboration, they should go and be turned into examples I think

pedrocamargo:
I don't think the goal is to rush, @janzill , but @felasouza approved the PR, so I went in for a deeper review (missed the WIP tag)

felasouza:
Hi @janzill,

I commited a few changes (mostly to not undermine performance for large
instance such as Chicago - which is indeed a bit slow anyway).

I harded coded a parts per partition on the path_based_assignment
parameter. My understanding is that I would need to retrieve from the
assign_spec. Should I change it or should I hold? Also I am answering
commenting on your points:

1 - The link class/struct can be defined externally. Not sure how I would
hold the derivatives and so on. You mean to use the link definitions and
use the definitions of bpr.pyx as oppose to the operations performed on
update_link_flows?

2 - this using the Cython implementation or boost? You mean the handling on
the path_based_assignment.py being transferred to the
trafficAssignment.cpp? Please guide me on that.

3 - sure. As I understand, the TrafficAssignmentCy would receive an
aequilibrae graph and AEquilibraE matrix directly?

I think the easier way to exchange numpy arrays as oppose to std vectors
(at least when the array sizes are fixed such as link flows/origin link
flows and such). For the paths, I am holding a
std::vector<std::vector<int>> for paths but should not be difficult to
exchange to numpy array of nodes and links. (I think the easier would be
like here
https://scipy.github.io/old-wiki/pages/C%2B%2B_Extensions_that_use_NumPy_arrays.html
)

Just give any hint and I can try to work on that.

Best regards,

Felipe

Em dom., 10 de jan. de 2021 às 00:13, Jan <notifications@github.com>
escreveu:

janzill:
Hi @felasouza,

I haven't really had any time to look at this further, sorry.

> My understanding is that I would need to retrieve from the assign_spec. Should I change it or should I hold? 

I think you can leave it as hard-coded for now, I'm happy to move it up once we want to play around with the parameter.

> 1 - The link class/struct can be defined externally. Not sure how I would hold the derivatives and so on. You mean to use the link definitions and use the definitions of bpr.pyx as oppose to the operations performed on update_link_flows?

Yes, exactly. From a maintenance stand-point having everything in one place is pretty important I think. Regarding derivatives of the vdfs, we do have these implemented as part of stepsize finding for biconjugate Frank-Wolfe, I actually started work on carrying this through and then realised I needed to think about this a bit more because of the cython/c++ mix. 

> 2 - this using the Cython implementation or boost? You mean the handling on the path_based_assignment.py being transferred to the trafficAssignment.cpp? Please guide me on that.

Pedro's cython code. I have to talk to Pedro about this, I'm not sure about how to efficiently do this because of the multi-threading implementation.

> 3 - sure. As I understand, the TrafficAssignmentCy would receive an aequilibrae graph and AEquilibraE matrix directly? I think the easier way to exchange numpy arrays as oppose to std vectors (at least when the array sizes are fixed such as link flows/origin link flows and such). For the paths, I am holding a std::vector<std::vector<int>> for paths but should not be difficult to exchange to numpy array of nodes and links. (I think the easier would be like here https://scipy.github.io/old-wiki/pages/C%2B%2B_Extensions_that_use_NumPy_arrays.html ) Just give any hint and I can try to work on that.

Agreed! I think we don't need to expose paths to cython for now, they are not needed during calculations. If we want these for post-processing or saving we could just export at the end. With regards to the actual implementation, I've started work on this but because of some of the slicing that's happening in aequilibrae we need to think about how to make these arrays c-contiguous.


Maybe it would be best to arrange a meeting between the three of us so we can flesh some of this out and see who does what, what do you think? @pedrocamargo and I are in the same time zone, you are about 16h behind us I think. Maybe early/mid next week before the TRB sessions start?

pedrocamargo:
Meeting would be a good idea, @janzill

felasouza:
Sounds good. Before the TRB session looks easier to me than for people in
Australia :-)

Em sex., 15 de jan. de 2021 às 03:31, Pedro Camargo <
notifications@github.com> escreveu:

PR #203, opened 2021-01-03
Changes network trigger structure to allow for disconnected centroids to exist

pedrocamargo:
- It changes triggers to allow centroids to stay in the network when no links are connected to it
- Adds network cleaning resource to remove regular disconnected nodes when opening and closing the network


aequilibrae/project/database_specification/triggers/network_triggers.sql
```diff
@@ -221,7 +223,13 @@ create trigger deleted_link after delete on links
     end;
 --#
 -- when moving OR creating a link, don't allow it to duplicate an existing link.
```
jamiecook, line 225: these comments don't line up with the code?

pedrocamargo: I had ignored the comments when updating the triggers. Good catch. Fixed

pedrocamargo:
@jamiecook & @arobrien . I tried to create a trigger to prevent the creation of nodes with nothing connected, but it ended up preventing the creation of nodes when adding new links (because of the sequence of triggers). Does any of you have a suggestion on that?

janzill (approved):
Looks good to me

PR #206, opened 2021-01-14
Tests package upload

pedrocamargo:
The release of the 0.7 version failed during upload, so now we can test it.


pedrocamargo:
Test PyPi also does not allow for uploading the same version over and over

PR #212, opened 2021-01-22
Refactoring Graph object based on Pandas

pedrocamargo:
It replaces the use of NumPy record Arrays with Pandas DataFrames


pedrocamargo:
Apparently, this breaks stuff really badly, but tests missed it and I'm still not sure what is going on

pedrocamargo:
Problems are resolved

aequilibrae/paths/graph.py
```diff
@@ -20,19 +20,25 @@ def __init__(self):
...
         self.date = str(datetime.now())
 
         self.description = "No description added so far"
 
         self.num_links = -1
         self.num_nodes = -1
         self.num_zones = -1
-        self.network = False  # This method will hold ALL information on the network
-        self.graph = False  # This method will hold an array with ALL fields in the graph.
+        self.network = pd.DataFrame([])  # This method will hold ALL information on the network
```
janzill, line 40: method -> variable?

aequilibrae/paths/graph.py
```diff
@@ -82,148 +86,6 @@ def default_types(self, tp: str):
         else:
             raise ValueError("It must be either a int or a float")
 
-    def create_from_geography(
```
janzill, line 85: Do you want to add a note along the lines of "has been removed in version ..."? Not sure if it would be helpful though.

janzill (approved):
Looks good to me.

aequilibrae/paths/graph.py
pedrocamargo: It was used mostly on testing. It was a REALLY hard method to use, so I am not really worried about removing it. Will put in the documentation of the next release.

PR #215, opened 2021-01-31
Incorporates network simplification into Skimming and assignment

pedrocamargo:
Performance gains for a large network (300k bi-direction links/ 6,000 centroids) have been over 63% for skimming


aequilibrae/paths/graph.py
```diff
@@ -103,25 +109,206 @@ def prepare_graph(self, centroids: np.ndarray) -> None:
...
+
+        nodes = np.hstack([self.network.a_node.values, self.network.b_node.values])
+        counts = np.bincount(nodes)
+        counts[self.centroids] = 999
+
+        truth = (counts == 2).astype(np.int)
+        link_edge = truth[self.network.a_node.values] + truth[self.network.b_node.values]
+        link_edge = self.network.link_id.values[link_edge == 1]
+
+        simplified_links = np.empty(self.network.link_id.max() + 1)
```
janzill, line 162: maybe `np.repeat(-1, self.network.link_id.max()+1)` instead of these l.162-164?

aequilibrae/paths/graph.py
```diff
@@ -103,25 +109,206 @@ def prepare_graph(self, centroids: np.ndarray) -> None:
...
+            b_node = self.network.b_node.values[lidx]
+            drc = self.network.direction.values[lidx]
+            n = a_node if counts[a_node] == 2 else b_node
+            first_node = b_node if counts[a_node] == 2 else a_node
+
+            ab_dir = 0 if (first_node == a_node and drc < 0) or (first_node == b_node and drc > 0) else ab_dir
+            ba_dir = 0 if (first_node == a_node and drc > 0) or (first_node == b_node and drc < 0) else ba_dir
+
+            while counts[n] == 2:
+                if simplified_links[pre_link] >= 0:
```
janzill, line 191: I like assert here, i.e. assert `simplified_links[pre_link] == -1, "your message"`. tomato, tomato.

janzill (approved):
Looks good to me. The network simplification method is hard to read I think, but there's probably not much that can be done about that. I approve, especially since the run time improvements are awesome!

aequilibrae/paths/results/assignment_results.py
```diff
@@ -104,11 +102,16 @@ def reset(self) -> None:
...
-            self.total_link_loads.fill(0)
+            self.link_loads.fill(0)
+            self.compact_link_loads.fill(0)
+            self.compact_total_link_loads.fill(0)
         else:
             raise ValueError("Exception: Assignment results object was not yet prepared/initialized")
 
     def __redim(self):
+        self.compact_link_loads = np.zeros((self.compact_links + 1, self.classes["number"]), self.__float_type)
+        self.compact_total_link_loads = np.zeros(self.compact_links, self.__float_type)
```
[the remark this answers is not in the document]
pedrocamargo: Yes.  It happens that some links get completely removed when the network is simplified (due to coding error, a link segment on a one-way road has the wrong direction), so instead of having to deal with this through conditions, I put one extra element in the compact link loads and refer the dropped links to it.
This ends up being the case for links only available for some classes and not others, which appear with a_node==b_node in the uncompressed graph but are completely removed from the simplified one.

aequilibrae/paths/graph.py
pedrocamargo: Good one. I am incorporating it

aequilibrae/paths/graph.py
pedrocamargo: This was actually a debug thing when I was implementing the code. removed it

aequilibrae/paths/results/assignment_results.py
```diff
@@ -104,11 +102,16 @@ def reset(self) -> None:
         if self.link_loads is not None:
             self.no_path.fill(0)
             self.link_loads.fill(0)
-            self.total_link_loads.fill(0)
+            self.link_loads.fill(0)
```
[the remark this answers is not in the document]
pedrocamargo: THat was actually a bug!

PR #218, opened 2021-02-05
Fixes bug with not being open to retrieve a matrix with names with upper case letters

### LINKED ISSUE #217: Fail to retrieve matrix when name has upper case letters
**Purpose of use**
Retrieving matrices for assignment

**Describe the bug**
The line of code below returns an error, despite the matrix existing
demand = proj_matrices.get_matrix('AM_omx')

**To Reproduce**
image

**Expected behavior**
N.A.

 - OS: Windows 64
 - Python version: 3.7.4
 - Version 0.7.1 dev

pedrocamargo:
There is still an edge case where somebody could try to have different matrix records with the same names but different upper/lower cases, which would break things. Will only address if it ever becomes an issue

close #217

### DIFF
--- a/docs/source/examples/plot_forecasting.py
+++ b/docs/source/examples/plot_forecasting.py
@@ -273,7 +273,7 @@ for function in ["power", "expo"]:
     gravity.apply()
 
     # We get the output matrix and save it to OMX too,
-    gravity.save_to_project(name=f"demand_{function}_model", file_name=f"demand_{function}_model.omx")
+    gravity.save_to_project(name=f"demand_{function}_modeled", file_name=f"demand_{function}_modeled.omx")
 
 # We put the aem files directly in the folder, so we update the database with them
 proj_matrices.update_database()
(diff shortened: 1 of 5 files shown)


docs/source/examples/plot_forecasting.py
```diff
@@ -285,18 +286,20 @@
...
+    "row_field": "origins",
+    "nan_as_zero": True,
+}
 
 ipf = Ipf(**args)
 ipf.fit()
 
-ipf.save_to_project(name='demand_ipf', file_name='demand_ipf.aem')
-ipf.save_to_project(name='demand_ipf_omx', file_name='demand_ipf.omx')
+ipf.save_to_project(name="demand_ipfd", file_name="demand_ipfd.aem")
```
jamiecook, line 301: why are these types of changes mixed in?

jamiecook (approved):
Looks fine - small change with lower but the presence of the ' -> " and filename changing makes me a bit suspicious.

docs/source/examples/plot_forecasting.py
pedrocamargo: Black.  My auto-commit hook would not allow me to do the commit without it.  And I tried!

PR #219, opened 2021-02-05
Matrix export

pedrocamargo:
Addresses matrix export issue reported on the GUI repo.

https://github.com/AequilibraE/AequilibraE-GUI/issues/171

### DIFF
--- a/aequilibrae/matrix/aequilibrae_matrix.py
+++ b/aequilibrae/matrix/aequilibrae_matrix.py
@@ -1,3 +1,4 @@
+from functools import reduce
 import importlib.util as iutil
 import os
 import tempfile
@@ -704,13 +705,14 @@ class AequilibraeMatrix(object):
             omx_export.close()
 
         elif file_extension == ".CSV":
-            def f(name): 
+
+            def f(name):
                 coo = coo_matrix(self.matrix[name])
-                data = {"row": self.index[coo.row], "column": self.index[coo.col], core: coo.data}
+                data = {"row": self.index[coo.row], "column": self.index[coo.col], name: coo.data}
                 return pd.DataFrame(data).set_index(["row", "column"])
-                
+
             dfs = [f(name) for name in self.names]
-            df = reduce(lambda a,b: a.join(b,how='outer'), dfs)     
+            df = reduce(lambda a, b: a.join(b, how="outer"), dfs)
             df.to_csv(output_name, index=True)
 
     def load(self, file_path: str):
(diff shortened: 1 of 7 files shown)


aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -702,26 +704,16 @@ def export(self, output_name: str, cores: List[str] = None):
...
-            self.computational_view(names)
+            df = None
+            for core in self.names:
+                coo = coo_matrix(self.matrix[core])
+                df2 = pd.DataFrame({"row": self.index[coo.row], "column": self.index[coo.col], core: coo.data})
+                df2.set_index(["row", "column"], inplace=True)
+                if df is None:
+                    df = df2
+                else:
+                    df = df.join(df2, how="outer")
```
jamiecook, line 715:
```suggestion
            def f(name): 
                coo = coo_matrix(self.matrix[name])
                data = {"row": self.index[coo.row], "column": self.index[coo.col], core: coo.data}
                return pd.DataFrame(data).set_index(["row", "column"])
                
            dfs = [f(name) for name in self.names]
            df = reduce(lambda a,b: a.join(b,how='outer'), dfs)     
```

tests/aequilibrae/project/test_matrices.py
```diff
@@ -29,90 +30,90 @@ def tearDown(self) -> None:
...
-        self.assertEqual(df.shape[0], self.curr.fetchone()[0], 'Returned the wrong number of matrices in the database')
+        self.curr.execute("select count(*) from Matrices")
+        self.assertEqual(df.shape[0], self.curr.fetchone()[0], "Returned the wrong number of matrices in the database")
 
-        self.assertEqual(df[df.status == 'file missing'].shape[0], 1, 'Wrong # of records for missing matrix files')
+        self.assertEqual(df[df.status == "file missing"].shape[0], 1, "Wrong # of records for missing matrix files")
 
     def __mat_count(self, should_have: int, error_message: str) -> None:
-        self.curr.execute('Select count(*) from Matrices;')
+        self.curr.execute("Select count(*) from Matrices;")
```
jamiecook, line 118: You should do one PR where you just run black over the entire codebase and then never worry about " vs ' again

pedrocamargo: Will do

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: Good one

janzill:
@pedrocamargo part of this is about internal handling of matrix names, right? I think making comparisons case-insensitive is a good idea, just needs a warning/note in the documentation. We might want to normalise things at some point to deal with non-ASCII letters, see for example https://stackoverflow.com/a/29247821 but I think this is good for now

janzill (approved):

PR #223, opened 2021-02-14
Fixes assignment with compressed graph and multiple classes

pedrocamargo:
The issue I was facing was that we were using the traffic classes' modes as dictionary keys.  Since the modes were derived from the graphs, things would go haywire when we had the same graph used for more than one class (Truck, in our case here).  I have replaced the dictionary keys with the classes' IDs, which are randomly generated using the **uuid** library and I still check if all classes have unique IDs.
Since I was at it, I introduced the multiplication of PCE at the beginning (right after the AoN), and just dividing the final result by it, as it reduces the chances of mixing up what are step directions (actual flows) Vs. congestion-causing flows for the purpose of speed computation (**with** PCEs).

### DIFF
--- a/aequilibrae/paths/linear_approximation.py
+++ b/aequilibrae/paths/linear_approximation.py
@@ -114,7 +114,7 @@ class LinearApproximation(WorkerThread):
         for c in self.traffic_classes:
             r = AssignmentResults()
             r.prepare(c.graph, c.matrix)
-            self.step_direction[c.mode] = r
+            self.step_direction[c.__id__] = r
 
         if self.algorithm in ["cfw", "bfw"]:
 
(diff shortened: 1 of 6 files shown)


janzill:
@pedrocamargo can you point me to the place you scale by PCEs? I cannot see it in this pr

aequilibrae/paths/linear_approximation.py
```diff
@@ -393,12 +400,18 @@ def execute(self):
...
-                    idx = c.graph.skim_fields.index(self.time_field)
-                    c.graph.skims[:, idx] = self.congested_time[:]
                 c._aon_results.reset()
+                if self.time_field not in c.graph.skim_fields:
+                    continue
+                idx = c.graph.skim_fields.index(self.time_field)
+                c.graph.skims[:, idx] = self.congested_time[:]
+
+        for c in self.traffic_classes:
+            c.results.link_loads /= c.pce
```
pedrocamargo, line 412: We revert the loads back to vehicles here

aequilibrae/paths/linear_approximation.py
```diff
@@ -323,21 +326,26 @@ def execute(self):
...
             aon_flows = []
-            for c in self.traffic_classes:
+            for c in self.traffic_classes:  # type: TrafficClass
+                aggregate_link_costs(self.congested_time, c.graph.compact_cost, c.results.crosswalk)
                 aon = allOrNothing(c.matrix, c.graph, c._aon_results)
                 if pyqt:
                     aon.assignment.connect(self.signal_handler)
                 aon.execute()
-                aon_flows.append(c._aon_results.total_link_loads * c.pce)
+                c._aon_results.link_loads *= c.pce
```
pedrocamargo, line 335: We apply the PCEs here

pedrocamargo:
> @pedrocamargo can you point me to the place you scale by PCEs? I cannot see it in this pr

Done. Two comments in the PR

aequilibrae/paths/linear_approximation.py
janzill: Thanks, that's what I was looking for!

janzill (approved):

PR #224, opened 2021-02-15
Janzill/correct conjugate direction

janzill:
For multi-class assignment, the second order linear approximation step direction finder did not include some cross-class terms. Specifically, the implementation of Eq. 18 in our TRB paper only used diagonal elements with respect to traffic classes. This PR includes off-diagonal elements, which will hopefully speed up convergence.

### DIFF
--- a/aequilibrae/paths/linear_approximation.py
+++ b/aequilibrae/paths/linear_approximation.py
@@ -147,13 +147,22 @@ class LinearApproximation(WorkerThread):
         # this needs more work
         numerator = 0.0
         denominator = 0.0
+        prev_dir_minus_current_sol = {}
+        aon_minus_current_sol = {}
+        aon_minus_prev_dir = {}
+
         for c in self.traffic_classes:
             stp_dir = self.step_direction[c.__id__]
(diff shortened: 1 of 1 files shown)


janzill:
@pedrocamargo do you mind running you multi-class example with this? Just want to ensure results are the same and I'd be interested to see if it matters in practice

pedrocamargo:
> @pedrocamargo do you mind running you multi-class example with this? Just want to ensure results are the same and I'd be interested to see if it matters in practice

Will do

pedrocamargo:
I tested for the large model we are working with and the result is this

image

I want to test on the Chicago TNTP instance before we merge, but it seems that we are requiring two iterations before the threshold, which was not the intention.

pedrocamargo:
> I tested for the large model we are working with and the result is this
> 
> 
> I want to test on the Chicago TNTP instance before we merge, but it seems that we are requiring two iterations before the threshold, which was not the intention.

The idea of Running Chicago is dumb... That is a single-class model....

pedrocamargo (approved):
The math is way beyond me and the results are solid, so I approve the changes. 

Some code cleaning to make it more readable in the future could be good, but that might be a good exercise for somebody looking to learn how BFW works.

PR #225, opened 2021-02-15
[WIP] Janzill/path saving

janzill:
adds path saving to assignment. Simple first implementation, probably terribly inefficient but easy to write. I like parquet/feather for this. PyArrow provides access to the underlying C++ library directly, which would let us do the writing without gil and make things fast. However, the compilation would have more dependencies and there are a few steps involved for the user to set it up. I'll do some testing with a larger network soon.


janzill:
too memory hungry and slow, only good for small test networks. See #232 for better imlplementation.

PR #226, opened 2021-02-16
[WIP] Bush based assignment

janzill:
@pedrocamargo @felasouza with all the changes happening in the new release and some funny business during merging, I decided to apply the latest changes to the current master branch, therefore this replaces #197 . Unfortunately, things are broken for my test case. I'm in the process of narrowing it down, but this might take a bit, especially since Pedro introduced network simplification along the way. There have been quite a few changes since I last looked at it, @felasouza was this latest version running for you?

Anyways, once this is sorted out the plan is as follows, as discussed earlier today:

1. Remove all duplicate calculations in the path-based C++ code in favour of existing implementations (e.g. congested link travel time, derivative of BPR, etc.). This could also go the other way, where we replace existing code with C++ bits, but it needs to be in one place.
2. Write tests and examples

This will get us to a first release version. We can then think about using a C++ QP solver without going back to python, about implementing the parallel version, and maybe most importantly about implementing this for multiple user classes. Did I forget anything?

### DIFF
--- a/.github/tests_linux.yml
+++ b/.github/tests_linux.yml
@@ -26,6 +26,7 @@ jobs:
         pip3 install pytest
         pip3 install spatialite pysqlite3-binary
         pip3 install pytest
+        pip3 install cvxopt
         sudo ln -s /usr/lib/x86_64-linux-gnu/mod_spatialite.so /usr/lib/x86_64-linux-gnu/mod_spatialite
 
     - name: Compile library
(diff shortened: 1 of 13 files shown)


janzill:
It's working now, but it's wrong. The problem is that I integrated things before Pedro implemented network simplification. The C++ code now uses the simplified network too, however the link cost calculations there do not. Due to the duplication of link cost calculations, we want to move that part out anyways and let the current cython code handle this. Calculations are correct there, so I'm not implementing the link cost calculation for the simplified network in C++, but will simply replace things.

PR #227, opened 2021-02-18
Adds upload to S3

pedrocamargo:


jamiecook (approved):
Looks legit man - getting some solid infrastructure together now.

PR #229, opened 2021-02-23
stepsize calculation fix

janzill:


janzill:
I think I need to use your black settings, these back and forth formatting changes are distracting

janzill:
there are two changes, one on line 329

janzill:
the other changes on on lines 435, now 460

jamiecook:
I've been saying that the use of black should just be baked in to the CI pipeline. 
Black shouldn't need any settings/config - except maybe line-length, I always just use

```
black --line-length=120 .
```

pedrocamargo:
> I've been saying that the use of black should just be baked in to the CI pipeline.
> Black shouldn't need any settings/config - except maybe line-length, I always just use
> 
> ```
> black --line-length=120 .
> ```

But at least this time It wasn't me!!! The issue here is that there is not a black config file in the project, so whatever the IDE is config with goes ahead.  I will see if I can prepare that config file and run black with it (shouldn't change much) when I can.

jamiecook:
what i've done is just add a linting stage that makes sure that command has been run.
Black has a check mode that will error if it would change any files

```
black --check --line-length 120 src-dir
```

put that into a linting stage in your ci pipeline and then make sure it passes before merging any PRs. Done

jamiecook:
image

But yeah - maybe adding that to the pyproject.toml will help other IDE's to get things consistent

janzill:
@pedrocamargo I cythonized the stepsize finder, and pulled out the non-stepsize dependent part (i.e. fixed cost contributions) on top. Makes a huge difference for the Birmingham example.

pedrocamargo:
Looks really good to me.  Do you mind if I commit to this same branch reverting the Black changes?  I will also fix on documentation.

janzill:
> Looks really good to me. Do you mind if I commit to this same branch reverting the Black changes? I will also fix on documentation.

Sure, go for it. Gotta remember to set my line length to 120 characters :)

janzill:
Note that we might want to think about the line search optimisation algorithm and its parameters as discussed earlier

pedrocamargo:
@ I merged the supernetwork work here already so I could make the fix for indexing the fixed cost

pedrocamargo (approved):
Looks good to me and it has been as tested as we could've without a reference instance.

PR #232, opened 2021-03-07
[WIP] Janzill/path saving nogil

janzill:


janzill:
hm, I think I was a bit too optimistic here, for real-world networks the number of files is very large with the current imlpementation. I will have to either switch to HDF5 and have variable-length arrays (could have one file per iteration per user class), or define a custom binary format.

pedrocamargo:
> hm, I think I was a bit too optimistic here, for real-world networks the number of files is very large with the current imlpementation. I will have to either switch to HDF5 and have variable-length arrays (could have one file per iteration per user class), or define a custom binary format.

Isn't HDF5 much less compressed?  Would a file per iteration make sense?  Or even a file per origin?

janzill:
yeah I thought about it a bit more and went with two feather (parquet is also implemented, but not exposed yet) files per iteration, class, and origin. One storing the path links for each destination sequentially, the other with index information to be able to split the paths by o-d pair. I think this is the easiest way. On my 8-thread laptop with 32GB an iteration of Birmingham with three classes without path saving takes about 2s, saving path brings this to about 7-8s. Each iteration has about 145MB of data.
This is all with gil, I let pyarrow handle writing, but I might be able to use their cython implementation to import C++ wrappers. But one thing at a time, I first need to dump the simplified network correspondence to disk.

janzill:
@pedrocamargo if you want to test on bigger instances, all you need is to call assig.set_save_path_files(True), using your traffic_assignment instance naming convention :)

pedrocamargo:
> @pedrocamargo if you want to test on bigger instances, all you need is to call assig.set_save_path_files(True), using your traffic_assignment instance naming convention :)

I will test it as soon as I fix the issue with the super networks (which is turning out to be a bit trickier than expected).

aequilibrae/paths/setup_Assignment.py
```diff
@@ -43,20 +44,30 @@
...
             "AoN",
             ["AoN.pyx"],
-            extra_compile_args=['-fopenmp'],
-            extra_link_args=['-fopenmp'],
-            include_dirs=[np.get_include()],
+            extra_compile_args=["-fopenmp", "-std=c++11", "-O3"],  # do we want -Ofast?
+            extra_link_args=["-lparquet", "-fopenmp"],
+            include_dirs=[np.get_include(), pa.get_include()],
+            language="c++",
+            runtime_library_dirs=["/mnt/c/Users/jan.zill/code/aequilibrae/.venv/lib/python3.7/site-packages/pyarrow/"],
```
pedrocamargo, line 42: Do we need this line?  Or to replace this with something generic?

aequilibrae/paths/setup_Assignment.py
```diff
@@ -43,20 +44,30 @@
         Extension(
             "AoN",
             ["AoN.pyx"],
-            extra_compile_args=['/openmp'],
-            extra_link_args=['/openmp'],
+            extra_compile_args=["/openmp", "/O2"],
+            extra_link_args=["/openmp"],
             include_dirs=[np.get_include()],
```
pedrocamargo: I added the Windows flag here

aequilibrae/paths/setup_Assignment.py
janzill: yes, done

pedrocamargo:
I am getting compilation errors on Windows 

warning: I don't know what to do with 'runtime_library_dirs': ['C:\\Users\\pedro\\.virtualenvs\\aequilibrae-ZVtI-VM1\\lib\\site-packages\\pyarrow', 'C:\\Users\\pedro\\.virtualenvs\\aequilibrae-ZVtI-VM1\\lib\\
site-packages\\pyarrow']
error: don't know how to set runtime library search path for MSVC


This seems to have the answer, but I don't quite understand what it does

https://github.com/Unidata/netcdf4-python/pull/461/files

janzill:
Looks like it does not include any arguments :) Dynamic loading is complicated enough on a single platform... see if it works without specifying runtime_library_dir.

janzill:
Didn't pay attention to windows so far, try now @pedrocamargo

pedrocamargo:
It does compile, but I get *DLL load failed: The specified module could not be found.*

janzill:
that's what including the runtime  library dir fixed for me on linux. I'd need to set it up on windows and look into it. My timing on an 8-thread machine for the Birmingham instance gave me the following results for BFW:
 1) w/o path file saving: 2.2s per iter
 2) cython write to feather: 8.6s
 3) c++ feather with gil: 15-16s
 4) c++ feather without gil: 6-7s
Option 2 would work for you, you only need to change the following things in AoN.pyx (the options are commented so I could switch quickly):
 - import path_file_saving instead of path_file_saving_nogil
 - comment lines 136 and 137, and uncomment lines 162 and 163

pedrocamargo:
> that's what including the runtime library dir fixed for me on linux. I'd need to set it up on windows and look into it. My timing on an 8-thread machine for the Birmingham instance gave me the following results for BFW:
> 
> 1. w/o path file saving: 2.2s per iter
> 2. cython write to feather: 8.6s
> 3. c++ feather with gil: 15-16s
> 4. c++ feather without gil: 6-7s
> 
> Option 2 would work for you, you only need to change the following things in AoN.pyx (the options are commented so I could switch quickly):
> 
> * import path_file_saving instead of path_file_saving_nogil (lines 26 and 27 need hashtag switched)
> * comment lines 136 and 137, and uncomment lines 162 and 163

janzill:
RuntimeError: Not supported on 32-bit Windows.
This is the remaining problem for windows checks above, for pyarrow builds. I need to look into it. 
For mac, I think all I need to do is tell clang to use c++11.

janzill:
Closing this as cython access avoids dll hell and windows packaging problems, and we take a marginal performance hit on the real world instances I tested. Superseded by #243

Issue #233, opened 2021-03-07
Super networks are inconsistent across different modes


pedrocamargo:
**Purpose of use**
Multi-class traffic assignment

**Describe the bug**
Links appear in different orders in graphs for different modes, but the IDs used to relate to compressed graphs appear in the same order, which is clearly wrong

**To Reproduce**
Generate graphs for different modes with different sets of links

**Expected behavior**
Same graph IDs corresponding to same links/directions

**Screenshots**
![image](https://user-images.githubusercontent.com/7377666/110232140-c184d880-7f67-11eb-9f17-4f91bea5ec68.png)

**Desktop (please complete the following information):**
 - OS: Win 10
 - Python version: 3.7.4
 - Version 0.7.1

PR #234, opened 2021-03-10
Introduces supernetwork graph ID across entire assignment

pedrocamargo:
Addresses issue #233 

@janzill , this should solve the problem, and it turns out to be rather elegant, so I am hoping this will work.

I will remove the WIP tag if the results for the generalized cost tests running tonight work accordingly

### DIFF
--- a/aequilibrae/paths/graph.py
+++ b/aequilibrae/paths/graph.py
@@ -124,6 +124,12 @@ class Graph(object):
 
         properties = self.__build_directed_graph(self.network, centroids)
         self.all_nodes, self.num_nodes, self.nodes_to_indices, self.fs, self.graph = properties
+
+        # We generate IDs that we KNOW will be constant across modes
+        self.graph.sort_values(by=['link_id', 'direction'], inplace=True)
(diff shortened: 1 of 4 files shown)


janzill:
@pedrocamargo can I clarify something. When we set up assignment, we define individual traffic classes, and each of these will have a graph, which is prepared for computation by a call to prepare_graph. Are these graphs necessarily the same in terms of the underlying network, i.e. will each link be present even if it has zero capacity for a class? If so, this PR looks good. If not, then assigning a supernetwork id independently with the implementation here won't work.

pedrocamargo:
> lasses, and each of these will have a graph, which is prepared for computation by a call to prepare_graph. Are these graphs necessarily the same in terms of the underlying network, i.e. will each link be present even if it has zero capacity for a class? If so, this PR looks good. If not, then assigning a supernetwork id independently with the implementation here won't work.

Yes, they are necessarily the same. The graph DataFrame inside the Graph object contains all the links in the network, but with a_node set to be equal to b_node for the links that are not for that mode.  It is a very inefficient way to build the Graph, but ensures consistency across all modes

janzill (approved):
looks good to me

PR #235, opened 2021-03-12
Documentation and examples for new features

pedrocamargo:
Already uploaded Sioux Falls Multi-Class, but there is still the following to be done:

 - [ ] Upload Birmingham Multi-class
 - [ ] Reference both new examples in the documentation
 - [ ] Write documentation for generalized cost in the traffic assignment page


pedrocamargo:
@jan, this is basically your branch with one new piece of documentation. So I am just going to merge it when it is done

PR #240, opened 2021-03-20
Addresses issue #238

### LINKED ISSUE #238: Network simplification failing
**Purpose of use**
Regular Traffic assignment

**Describe the bug**
The network simplification step fails when building graphs. It happens any time the highest node_id in the network is in the middle of a section that will be simplified out

**To Reproduce**
Create a network with links/nodes as described above

 - OS: Windows 10
 - Python version. 3.7
 - Version: 0.7.2 (dev)

pedrocamargo:
Closes #238

### DIFF
--- a/aequilibrae/paths/graph.py
+++ b/aequilibrae/paths/graph.py
@@ -143,22 +143,23 @@ class Graph(object):
 
         nodes = np.hstack([self.network.a_node.values, self.network.b_node.values])
         links = np.hstack([self.network.link_id.values, self.network.link_id.values])
+        counts = np.bincount(nodes)
 
         idx = np.argsort(nodes)
         all_nodes = nodes[idx]
         all_links = links[idx]
-        links_index = np.empty(all_nodes.max() + 1, np.int64)
+        links_index = np.empty(all_nodes.max() + 2, np.int64)
         links_index.fill(-1)
-        nlist = np.arange(all_nodes.max())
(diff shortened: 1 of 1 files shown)


aequilibrae/paths/graph.py
```diff
@@ -143,22 +143,23 @@ def __build_compressed_graph(self):
...
         links_index.fill(-1)
-        nlist = np.arange(all_nodes.max())
+        nlist = np.arange(all_nodes.max() + 2)
 
         y, x, _ = np.intersect1d(all_nodes, nlist, assume_unique=False, return_indices=True)
         links_index[y] = x[:]
         links_index[-1] = all_links.shape[0]
-        for i in range(all_nodes.max(), 1, -1):
+
+        for i in range(all_nodes.max() + 1, 1, -1):
```
janzill, line 159: The index will go down to 2, so links_index[0] will never be touched. Is this intended or should it be `for i in range(all_nodes.max() + 1, 0, -1):`

janzill (commented):
@pedrocamargo just a clarification on setting the links_index values

aequilibrae/paths/graph.py
pedrocamargo: That does not make much difference because that is already filled with something different than -1. But you are actually right that we need it, as it is not a great practice to have it this way.

janzill (approved):

PR #243, opened 2021-03-23
[WIP] select link analysis via path files

janzill:
Path file saving is now in #253, this bit here is for select link analysis via these path files. This is slower than an implementation during assignment, however it has the advantage that it works for any link(s) as long as path files have been stored during assignment.

The main reason to go down this path is that path file saving is available and therefore this implementation is a lot less work.


janzill:
@pedrocamargo this is faster than I thought, pretty much nogil c++ speed, because pyarrow is great, and because going from std::vector to a numpy array can be done without copying data. I'm happy with performance and compression. I've left writing to parquet exposed as an option. It's using snappy compression by default and is slower and larger than feather. My tests also indicate that using gzip compression doesn't make it better than l4z in our case, so I'd always go with feather and I've made that the standard format. The only option that's missing is making the base path configurable such that we can pass in a time period for example.

aequilibrae/paths/path_file_saving.pyx
```diff
@@ -0,0 +1,74 @@
...
+    cdef vector[long long] size_of_path_arrays
+    cdef long long* temp_data
+    cdef long long* temp_data_ind
+
+    cdef np.npy_intp dims[1]
+    cdef np.ndarray[np.longlong_t, ndim=1] numpy_array
+    cdef np.npy_intp dims_ind[1]
+    cdef np.ndarray[np.longlong_t, ndim=1] numpy_array_ind
+
+    for node in range(zones):
```
pedrocamargo, line 41: This section would be inside a 

`with nogil:`

block

aequilibrae/paths/linear_approximation.py
```diff
@@ -325,6 +329,15 @@ def execute(self):
                 # cost = c.fixed_cost / c.vot + self.congested_time #  now only once
                 cost = c.fixed_cost + self.congested_time
                 aggregate_link_costs(cost, c.graph.compact_cost, c.results.crosswalk)
+
+                if c._aon_results.save_path_file:
+                    # TODO Jan: make base dir user configurable
+                    pth = os.environ.get(ENVIRON_VAR)
+                    path_base_dir = os.path.join(pth, "path_files", self.procedure_id, f"iter{self.iter}")
+                    # TODO Jan: what identifier do we use for the class?
+                    c._aon_results.path_file_dir = os.path.join(path_base_dir, f"path_c{c.mode}")
```
pedrocamargo, line 338:
```suggestion
                    c._aon_results.path_file_dir = os.path.join(path_base_dir, f"path_c{c.mode}_{c.__id__}")
```

Using mode only leads us to that issue we had before of multiple classes using the same road mode. We could think about another way of differentiating the classes (tying to the matrices, maybe), but for now, we should have the ID there

aequilibrae/paths/path_file_saving.pyx
janzill: yes, good catch

aequilibrae/paths/linear_approximation.py
janzill: true, forgot about that

pedrocamargo:
@janzill, I committed some stuff here that I worked on, so let's keep this around so we can separate things on different PRs.

janzill:
@pedrocamargo do you remember what bits you added here? Only those two refactoring ones? I've cleaned up the path file saving PR, there's so much stuff that it'd be quicker to just branch off of that if your only changes were those two quick-refactoring commits. Are you ok with me porting these over to janzill/select_link_analysis?

janzill:
@pedrocamargo I cherry picked those commits, closing this and re-opening based on new path file saving.

PR #245, opened 2021-03-26
Adds missing information to assignment results/report

pedrocamargo:
Adds information on PCEs, VoTs, Fixed costs and their multipliers to the assignment report

### DIFF
--- a/aequilibrae/paths/traffic_assignment.py
+++ b/aequilibrae/paths/traffic_assignment.py
@@ -481,19 +481,24 @@ class TrafficAssignment(object):
             *info* (:obj:`dict`): Pandas dataframe with all the assignment results indexed on link_id
         """
 
-        classes = {'matrix_totals': {}}
+        classes = {}
+
         for cls in self.classes:
+            uclass = {}
+
             if len(cls.matrix.view_names) == 1:
-                classes['matrix_totals'][cls.graph.mode] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for nm in
-                                                            cls.matrix.view_names}
+                uclass['matrix_totals'] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for nm in
+                                           cls.matrix.view_names}
             else:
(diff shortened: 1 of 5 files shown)


aequilibrae/paths/traffic_assignment.py
```diff
@@ -481,14 +481,19 @@ def info(self) -> dict:
             *info* (:obj:`dict`): Pandas dataframe with all the assignment results indexed on link_id
         """
 
-        classes = {}
+        classes = {'matrix_totals': {}}
         for cls in self.classes:
             if len(cls.matrix.view_names) == 1:
-                classes[cls.graph.mode] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for nm in cls.matrix.view_names}
+                classes['matrix_totals'][cls.graph.mode] = {nm: np.sum(cls.matrix.matrix_view[:, :]) for nm in
```
pedrocamargo, line 487: This suffers from the same issue we had inside the liner_approximation class.  I'll address it before we merge it

aequilibrae/paths/traffic_assignment.py
```diff
@@ -482,13 +482,23 @@ def info(self) -> dict:
...
-                }
+                uclass['matrix_totals'] = {nm: np.sum(cls.matrix.matrix_view[:, :, i]) for i, nm in
+                                           enumerate(cls.matrix.view_names)}
+            uclass['network mode'] = cls.graph.mode
+            uclass['Value-of-time'] = cls.vot
+            uclass['PCE'] = cls.pce
+            if cls.fixed_cost_field:
+                uclass['Fixed cost field'] = cls.fixed_cost_field
+                uclass['Fixed cost multiplier'] = cls.fc_multiplier
+            classes[cls.__id__] = uclass
```
janzill, line 501: Should we make sure that class names are unique and raise a helpful error otherwise? Or is this already part of the code and therefore does not show up in the changes here?

janzill (commented):
User-defined class names are a good idea I think, just a question with regards to enforcement   of uniqueness

PR #248, opened 2021-04-08
fixes skim creation of simplified graph

### LINKED ISSUE #247: Error during 'set_skimming' process
I have been testing AequilibraE in Google Colab, it seems to me like a good way to go around the process of installing the components in Windows and also its a very portable solution. I managed to get everything working for the examples in your website, the Sioux Falls projects runs great.

I created my network with the QGIS package (it is not a huge network, it has a little over 3000 links)  and I got it to run correctly there, all the way to the assignment of my demand. I'm very happy about that.

I also went through the process of updating my GIS network to the 0.7 version as you told me but I think there's probably something wrong with it.

It loads in Python (I can see the link and node data), it creates the graphs, but when I try to define the skimming (graph.set_skimming) it gives an error:
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-6-483409e42323> in <module>()

---> 10 graph.set_skimming(['free_flow_time', 'distance'])

/usr/local/lib/python3.7/dist-packages/aequilibrae/paths/graph.py in set_skimming(self, skim_fields)
    473         df = self.__graph_groupby.sum()[skim_fields].reset_index()
    474         for i, skm in enumerate(skim_fields):
--> 475             self.compact_skims[df.index.values, i] = df[skm].values[:].astype(self.__float_type)
    476
    477         self.skims = np.zeros((self.num_links, len(skim_fields) + 1), self.__float_type)

IndexError: index 2634 is out of bounds for axis 0 with size 2634

 - OS: Google Colab (Linux)
 - Python = 3.7
 - Version ?
 - 
[network_v07.zip](https://github.com/AequilibraE/aequilibrae/files/6265155/network_v07.zip)

janzill:
fixes #247

### DIFF
--- a/aequilibrae/paths/graph.py
+++ b/aequilibrae/paths/graph.py
@@ -126,9 +126,9 @@ class Graph(object):
         self.all_nodes, self.num_nodes, self.nodes_to_indices, self.fs, self.graph = properties
 
         # We generate IDs that we KNOW will be constant across modes
-        self.graph.sort_values(by=['link_id', 'direction'], inplace=True)
+        self.graph.sort_values(by=["link_id", "direction"], inplace=True)
         self.graph.loc[:, "__supernet_id__"] = np.arange(self.graph.shape[0]).astype(self.__integer_type)
-        self.graph.sort_values(by=['a_node', 'b_node'], inplace=True)
+        self.graph.sort_values(by=["a_node", "b_node"], inplace=True)
 
         self.num_links = self.graph.shape[0]
         self.__build_derived_properties()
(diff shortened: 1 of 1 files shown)


janzill:
@pedrocamargo it looks like you add links with a_node==b_node after network simplification and set the id to compact_num_links+1, so the groupby object has one element more than the simplified graph.

pedrocamargo (approved):

jamiecook:
wouldn't it be better to not add links that go from a->a ? 
also just wondering how you can know that [:-1] will catch all such cases...

pedrocamargo:
The thing is that we set node_b = node_a for all links that are not part of that particular graph, and are not in the graph used for computing paths.  This is done for managing super networks in multi-class assignments, as that procedure guarantees the order of the links in the graphs for all modes/classes.  
The solution fixes it because the +1 that was introduced is for the Forward-Star that indexes links in the graph with everything.

PR #253, opened 2021-05-23
janzill/pathfilesaving

janzill:
Breaks #243 into two pieces because it's become a monster over time. This here is for the actual path file saving.


janzill:
hey @pedrocamargo this is just the path file saving bit, as discussed. We still need two things: tests and a reader. The latter is important because we write two files, one a continuous set of links for a given origin to all destinations, and one with indexes that identify which bits are for which origin. I want to add that functionality as part of the select link analysis code, and if people are keen to use the files before we can share a code snippet showing how to extract a path for a single o-d.

aequilibrae/paths/AoN.pyx
```diff
@@ -87,6 +92,25 @@ def one_to_all(origin, matrix, graph, result, aux_result, curr_thread):
     cdef double [:, :] node_load_view = aux_result.temp_node_loads[:, :, curr_thread]
     cdef long long [:] b_nodes_view = aux_result.temp_b_nodes[:, curr_thread]
 
+    # path saving file paths
+    cdef string path_file_base
+    cdef string path_index_file_base
+    cdef bool save_paths = False
+    cdef bool write_feather = True
+    if result.save_path_file == True:
```
jamiecook, line 100: if True == True: ??

janzill: result.save_path_file is a variable that defaults to false

janzill: did you confuse it with save_paths just above, or did I miss your point?

janzill: also, save_paths does not seem to be used anymore, so should be removed

jamiecook: missed the point i think

```suggestion
    if result.save_path_file:
```

janzill: indeed I did, didn't even notice :)

janzill:
@pedrocamargo tests are in, this is ready for review now

janzill:
@pedrocamargo I decided to add basic path file parsing functionality here. It's for zone indexes (the inverse of graph.compact_nodes_to_indices) and for compact link ids for now.

aequilibrae/paths/AoN.pyx
```diff
@@ -12,20 +12,25 @@
...
 cimport numpy as np
+from libcpp cimport bool
 
 # include 'parameters.pxi'
 include 'basic_path_finding.pyx'
 include 'bpr.pyx'
 include 'conical.pyx'
 include 'parallel_numpy.pyx'
+include 'path_file_saving.pyx'
+# include 'path_file_saving_nogil.pyx'
```
pedrocamargo, line 27: Do we have this commented file?

pedrocamargo:
We will have to make a release before we merge this one, as Pyarrow will break AequilibraE for QGIS on Windows.  The solution for that integration will be to make PyArrow optional for AequilirbaE throughout, but we don;t need to worry about that for now

aequilibrae/paths/linear_approximation.py
```diff
@@ -359,6 +363,22 @@ def execute(self):
...
                 cost = c.fixed_cost + self.congested_time
                 aggregate_link_costs(cost, c.graph.compact_cost, c.results.crosswalk)
+
+                if c._aon_results.save_path_file:
+                    # TODO (Jan 18/4/21): make base dir user configurable
+                    pth = os.environ.get(ENVIRON_VAR)
+                    path_base_dir = os.path.join(pth, "path_files", self.procedure_id, f"iter{self.iter}")
+                    c._aon_results.path_file_dir = os.path.join(path_base_dir, f"path_c{c.mode}_{c.__id__}")
+                    Path(c._aon_results.path_file_dir).mkdir(parents=True, exist_ok=True)
+                    if self.iter == 1:
```
pedrocamargo, line 373: Is it here the place, or in the _execute_ of traffic assignment before the assignment actually starts?

aequilibrae/paths/linear_approximation.py
```diff
@@ -445,6 +465,9 @@ def execute(self):
             c.results.link_loads /= c.pce
             c.results.total_flows()
 
+        # TODO (Jan 18/4/21): Do we want to blob store path files (by iteration, class, origin, destination) in sqlite?
```
pedrocamargo, line 468: Can we blob in Feather inside SQLite in the end?  I don't remember where that conversation went.

aequilibrae/paths/traffic_assignment.py
```diff
@@ -295,6 +299,37 @@ def set_cores(self, cores: int) -> None:
             c.results.set_cores(cores)
             c._aon_results.set_cores(cores)
 
+    def set_save_path_files(self, save_it: bool) -> None:
```
pedrocamargo, line 302: What if we add a class after setting it to save path?

pedrocamargo: Maybe a list of traffic classes should be in the instantiation of the TrafficAssignment class?

tests/aequilibrae/paths/test_assignment_results.py
```diff
@@ -23,13 +23,15 @@ def test_set_cores(self):
...
         self.assertEqual(a.cores, max(1, mp.cpu_count() - 2))
 
-    #
+    def test_set_save_path_file(self):
+        a = AssignmentResults()
+
+        # Never save by default
+        self.assertEqual(a.save_path_file, False)
+
     # def test_set_critical_links(self):
```
pedrocamargo, line 32: Are these tests coming?

pedrocamargo: Of course not!  Old stuff I left there!

aequilibrae/paths/linear_approximation.py
pedrocamargo: Worth trying? https://stackoverflow.com/a/58980160/1480643

aequilibrae/paths/graph.py
```diff
@@ -630,3 +630,6 @@ def __determine_types__(self, new_type, current_type):
         else:
             raise ValueError("WRONG TYPE OR NULL VALUE")
         return def_type
+
+    def save_compressed_correspondence(self, path):
+        self.graph[["link_id", "__supernet_id__", "__compressed_id__"]].to_feather(path)
```
pedrocamargo, line 635: Don't we need link direction as well, since the compressed graph is directional?

aequilibrae/paths/linear_approximation.py
pedrocamargo: Some comments on the design from of sqlite files with blobs from the creator of SQLite himself
http://sqlite.1065341.n5.nabble.com/Effect-of-blobs-on-performance-tp19559p19571.html

aequilibrae/paths/results/assignment_results.py
```diff
@@ -51,6 +51,11 @@ def __init__(self):
         self.lids = None
         self.direcs = None
 
+        # save path files. Need extra metadata for file paths
+        self.save_path_file = False
+        self.path_file_dir = None
+        self.write_feather = True  # we use feather as default, parquet is slower but with better compression
```
pedrocamargo, line 57: Is the distinction between parquet and Feather something we need to know when we will read the data?  If so, we need to include this in the Assignment report (the one where we write the betas and convergence report to)

aequilibrae/paths/assignment_paths.py
```diff
@@ -0,0 +1,119 @@
...
+# for zones, we need to do the inverse of graph.compact_nodes_to_indices
+
+
+class TrafficClassIdentifier(object):
+    def __init__(self, name: str, id: str):
+        self.name = name
+        self.id = id
+
+
+class AssignmentResultsTable(object):
```
pedrocamargo, line 27: I think some documentation is needed here

pedrocamargo: Or not... If it is used by another class and not meant to be used directly... My bad...

pedrocamargo (commented):
My biggest concern is with the saving of the correspondence, as I am pretty sure we will need the direction of the link there as well

aequilibrae/paths/AoN.pyx
janzill: no I removed all c++ code

aequilibrae/paths/linear_approximation.py
janzill: good question, I like having it in one place because the code is pretty monolithic already and hard to read, but the logic suggests doing this once before and then not checking for iteration numbers. I think I will try to make this more readable when thinking about the blob store.

aequilibrae/paths/assignment_paths.py
janzill: I think it would be good to have this available and documented for usability, but I'd say this is out-of-scope for this PR

aequilibrae/paths/graph.py
janzill: saving full graph and nodes_to_indices now

janzill:
> My biggest concern is with the saving of the correspondence, as I am pretty sure we will need the direction of the link there as well

as just discussed we don't need it strictly speaking, but it makes our life easier and in fact storing the entire graph on disk is not a problem and helpful for certain use cases so we will do that

aequilibrae/paths/linear_approximation.py
janzill: as just discussed this is out of scope, we'll make it an enhancement issue

aequilibrae/paths/linear_approximation.py
janzill: I've moved this into a little method and kept the if iter==1 statement in there, I like that it's self contained but let me know if you disagree

aequilibrae/paths/results/assignment_results.py
janzill: good point, I'll add it. At this point both formats are implemented so I'll just leave them, but I think we could also just remove parquet and only support feather.

aequilibrae/paths/traffic_assignment.py
janzill: it's really a convenience method that applies it to a list of current classes. either you add them on instantiation and then make them fixed, or you keep a flag that then applies this to each class you add.

janzill: that's really a user question so I am happy with any consistent definition as long as we make it specific

janzill:
@pedrocamargo requesting re-review please

pedrocamargo (approved):

PR #255, opened 2021-05-25
Fix volume over capacity calculation for assignment results

barisdemirdelen:
When calculating volume over capacity for assignment results, the calculation is made on inconsistent indices, which results in wrong voc values for the resuls.

This pull request ensures voc calculation is made with the correct indices.

### DIFF
--- a/aequilibrae/paths/traffic_assignment.py
+++ b/aequilibrae/paths/traffic_assignment.py
@@ -409,7 +409,7 @@ class TrafficAssignment(object):
         res1 = assig_results[0]
 
         tot_flow = self.assignment.fw_total_flow[idx]
-        voc = self.assignment.fw_total_flow / self.capacity[idx]
+        voc = tot_flow / self.capacity[idx]
         congested_time = self.congested_time[idx]
         free_flow_tt = self.free_flow_tt[idx]
 


pedrocamargo (approved):

aequilibrae/paths/traffic_assignment.py
```diff
@@ -409,7 +409,7 @@ def results(self) -> pd.DataFrame:
         res1 = assig_results[0]
 
         tot_flow = self.assignment.fw_total_flow[idx]
```
jamiecook, line 411: @pedrocamargo ..... it looks like the intention was always to use the tot_flow variable in the manner that @barisdemirdelen has fixed.

I say looks like as the variable is otherwise unused - which should be a flake8 violation.... I thought you had flake8 setup as a linting stage (but i can't seem to see it in the re-shuffle to use github workflows) - I would have expected it to catch something like this.

jamiecook: Sorry - that's my bad - the variable is used... my highlighting was just broken :/

jamiecook: nice catch @barisdemirdelen

PR #256, opened 2021-05-25
Fix shortest path calculation when the predecessor of destination node is node index 0

barisdemirdelen:
In the shortest path computation, predecessors_view array has the predecessor node indices or -1 for no path. The intention with the if statements in this pull request is to check if there is a path to destination.

Unfortunately in the cases where the predecessor of destination node is node with index 0, path computation thinks there is no path to the destination node.

This pull request fixes this issue by checking for >=0 instead of >0 for the predecessor

### DIFF
--- a/aequilibrae/paths/AoN.pyx
+++ b/aequilibrae/paths/AoN.pyx
@@ -228,7 +228,7 @@ def path_computation(origin, destination, graph, results):
                                     b_nodes_view,
                                     original_b_nodes_view)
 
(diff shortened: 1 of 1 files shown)


pedrocamargo:
@janzill, can you check this one?

janzill:
This makes sense, -1 is indeed what is being used for disconnections in the predecessor matrix. If 0 is a valid id, this should be included.

janzill (approved):

PR #257, opened 2021-05-25
Fix shortest path calculation with blocking centroids

barisdemirdelen:
Shortest path calculation with blocking centroids fails. This is because of blocking_centroid_flows() method in basic_path_finding.pyx

centroids variable contains the number of centroids in the graph. This method as is blocks and unblocks one too many node, which means  if the graph is set to block the centroids, the first non centroid node of the graph is always blocked. If a shortest path would contain this node, then we either get no path or wrong path as a result.

### DIFF
--- a/aequilibrae/paths/basic_path_finding.pyx
+++ b/aequilibrae/paths/basic_path_finding.pyx
@@ -168,10 +168,10 @@ cdef void blocking_centroid_flows(int action,
     cdef long long i
 
     if action == 1: # We are unblocking
-        for i in range(fs[centroids + 1]):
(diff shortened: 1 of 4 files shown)


pedrocamargo (approved):

barisdemirdelen:
I have some tests for this but not sure on where to put them. Let me try something and if you don't like it we can discuss

pedrocamargo:
> ot sure on where to put them. Let me try something and if you don't like it we can discuss

That would be great if you could include it

barisdemirdelen:
I have added my tests for this.

janzill (approved):

PR #258, opened 2021-05-29
fixes cfw stepdirection bug

janzill:
@pedrocamargo found a bug while working on tests, I'm a bit confused why this passed tests on github, but I haven't looked into it. Anyways, this makes the CFW step direction calculation consistent with the maths in our TRB paper, specifically Eq.10.


janzill:
@pedrocamargo as discussed on the phone just now, I'll merge this

pedrocamargo:
Thanks, man!!!!  My bad!!!

PR #262, opened 2021-06-18
Adds tests to make sure computing paths behaves properly on disconnected networks

pedrocamargo:
Tests two methods for disconnecting, directly in the network and in the graphs


janzill (approved):
looks good to me

PR #265, opened 2021-07-31
Adds Python 3.9 support

pedrocamargo:
Pytables (a requirement for OpenMatrix ) is not available for Python 3.9, so we need to change some stuff around to make it work

### DIFF
--- a/.github/workflows/test_windows.yml
+++ b/.github/workflows/test_windows.yml
@@ -46,7 +46,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install setuptools wheel
         pip install -r requirements.txt
-        pip install pytest, openmatrix
+        pip install pytest openmatrix
 
     - name: Download spatialite
       run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/test_windows.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [ '3.6', '3.7', '3.8' ]
```
janzill, line 33: @pedrocamargo no 3.9 on windows? Also, spacing is inconsistent

janzill (approved):
windows question, looks good otherwise

.github/workflows/test_windows.yml
pedrocamargo: Testing cannot be done because PyTables is not available for Python 3.9 yet (that's the HDF5 library used by OpenMatrix). In practice people can use by downloading the binaries from Chris Gohlke, but nothing we should put in here...

PR #267, opened 2021-08-25
Pedro/documentation

pedrocamargo:


pedrocamargo:
@janzill , the test for path saving is failing. Can you take a look at that since you have written that entire code?

janzill:
for reference, @pedrocamargo fixed this in #271 because I dropped the ball on this

PR #280, opened 2021-11-08
QGIS adjustments

pedrocamargo:
Incorporates changes required for integration with the QGIS GUI.

It also incorporates a few API changes and rationalizations that I wanted to get out of the way.

This is the last PR before the next release (Let's join Develop into master and do a new release after this one is merged)


janzill (approved):
looks good to me, always nice to remove code duplication

PR #287, opened 2021-12-10
Fixes inrets tests

pedrocamargo:
Fixes the Inrets test. There is a kink in the VDF, so the monotonicity of its derivative is not true around the vdf kink.


pedrocamargo:
@janzill , Only CFW is failing now...

pedrocamargo:
@janzill , can we move forward with this one and leave CFW for later?

janzill (approved):

PR #288, opened 2021-12-10
Two small changes around the edges

pedrocamargo:
Adds the obvious field vot to the modes table and creates and adds a script for compiling that may be useful during development

### DIFF
--- a/aequilibrae/project/database_specification/tables/modes.sql
+++ b/aequilibrae/project/database_specification/tables/modes.sql
@@ -1,7 +1,8 @@
 CREATE TABLE if not exists modes (mode_name   VARCHAR UNIQUE NOT NULL,
                                   mode_id     VARCHAR UNIQUE NOT NULL       PRIMARY KEY,
                                   description VARCHAR,
-                                  pce         NUMERIC        NOT NULL DEFAULT 0,
+                                  pce         NUMERIC        NOT NULL DEFAULT 1.0,
+                                  vot         NUMERIC        NOT NULL DEFAULT 0,
                                   alpha       NUMERIC,
                                   beta        NUMERIC,
                                   gamma       NUMERIC,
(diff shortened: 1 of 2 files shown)


pedrocamargo:
@jamiecook , don't worry about the test failures. Those are known issues elsewhere in the development branch

tests/compile.py
```diff
@@ -0,0 +1,8 @@
+from os.path import join, dirname
+import os
+import subprocess
+
+pth = join(dirname(dirname(__file__)), 'aequilibrae', 'paths')
+
+os.chdir(pth)
+subprocess.Popen("python setup_Assignment.py build_ext --inplace", shell=True, stdout=subprocess.PIPE).stdout.read()
```
jamiecook, line 8: Is it worth using a substitution here

```suggestion
subprocess.Popen(f"{sys.executable} setup_Assignment.py build_ext --inplace", shell=True, stdout=subprocess.PIPE).stdout.read()
```

to make sure the right python is called.

aequilibrae/project/database_specification/tables/modes.sql
```diff
@@ -1,7 +1,8 @@
 CREATE TABLE if not exists modes (mode_name   VARCHAR UNIQUE NOT NULL,
                                   mode_id     VARCHAR UNIQUE NOT NULL       PRIMARY KEY,
                                   description VARCHAR,
-                                  pce         NUMERIC        NOT NULL DEFAULT 0,
+                                  pce         NUMERIC        NOT NULL DEFAULT 1.0,
```
jamiecook, line 4: This default changed from 0 -> 1 ... intentional?

jamiecook: Guessing it probably is

jamiecook (approved):

PR #292, opened 2022-01-16
Pedro/fixes testing

pedrocamargo:
Binaries for PyTables are finally available for Python 3.9 & 3.10 on Windows, so we can test those versions fully now.

We also add full testing for MacOS & Ubuntu


janzill:
looks like macos is using clang instead of gcc

janzill:
do we want to pull out testing on macos into its own issue and include newer python versions and linux now?

pedrocamargo:
@janzill , I think that would be the right way to go, I think.

Issue #284, opened 2021-11-18
CFW regression on Develop branch


pedrocamargo:
**Purpose of use**
Development

**Describe the bug**
CFW is no longer working, and is causing tests to fail

**To Reproduce**
See https://github.com/AequilibraE/aequilibrae/pull/283

janzill:
To summarise the discussion in #283, whitespace changes in bpr.pyx led to the bug. Changing it back slightly altered traffic assignment test results, see #293. I'm removing the labels here (bug and critical) because they do not apply anymore but I'm keeping this issue open for now because I want to understand what is happening. It could be that very small changes in results (numerical precision) lead to larger differences in the step finding routine.

janzill:
Things to look at: 

- making sure we use double precision in calculations everywhere
- compare generated c++ code before and after white-space changes in bpr.pyx
- investigate step finding optimisation parameters

PR #293, opened 2022-01-16
fixes traffic assignment tests

janzill:
@pedrocamargo I don't think comparing rgap for converged bfw and cfw is a valid test, the only thing we can guarantee is that both are less than the specified target. We can, however, compare the number of iterations between tests.


janzill:
Note that the rgap test used to pass before the bpr changes and I don't know why it doesn't anymore. Like I said, I don't think the test is valid but I'd still like to understand where the change is coming from. I'll track this at #284.

pedrocamargo (approved):

Issue #285, opened 2021-12-03
Error when trying to add centroid


daniloebbinghaus:
Basically, it seems that when we have a node previously located at the geometric center of a zone, the method "add_centroid(None)" doesn't create a centroid and returns the following error (image below):

![add_centroid_bug](https://user-images.githubusercontent.com/91709021/144677388-c76f64b9-9159-4409-8ed0-34aa58ccab75.png)

pedrocamargo:
@daniloebbinghaus , do you have that query/code that allowed you to get around this?

daniloebbinghaus:
Yes, @pedrocamargo. The code is here:

```ruby
sql = """SELECT count(*) FROM nodes
            WHERE  nodes.geometry = GeomFromWKB(?, 4326) AND
         nodes.ROWID IN (
          SELECT ROWID FROM SpatialIndex WHERE f_table_name = 'nodes' AND
          search_frame = GeomFromWKB(?, 4326))
      """

zones_list = [x[0] for x in proj.conn.execute("select zone_id from zones where zone_id not in (select node_id from nodes)")]

for z_id in zones_list:
    zone = zoning.get(z_id)
    centroid = zone.geometry.centroid

    while True:
        test_list = proj.conn.execute(sql, [centroid.wkb, centroid.wkb]).fetchall()
        if test_list[0][0] == 0:
            break

        centroid = Point(zone.geometry.centroid.x + random.random()*0.0003, zone.geometry.centroid.y + random.random()*0.0003)
        
    zone.add_centroid(centroid)
```

So, adding randomly shifted centroids to all zones with nodes located at their very geometric centers ended up working as a workaround. I had to use that sql query to check if there was already a node located at the same position as the centroid.

PR #297, opened 2022-02-02
Improves robustness in the creation of centroids

pedrocamargo:
Closes issue #285

### DIFF
--- a/aequilibrae/project/zone.py
+++ b/aequilibrae/project/zone.py
@@ -1,3 +1,4 @@
+import random
 from sqlite3 import Connection
 from warnings import warn
 from shapely.geometry import Point, MultiPolygon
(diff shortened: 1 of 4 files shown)


aequilibrae/project/zone.py
```diff
@@ -75,8 +82,22 @@ def add_centroid(self, point: Point) -> None:
...
+                          nodes.ROWID IN (
+                           SELECT ROWID FROM SpatialIndex WHERE f_table_name = 'nodes' AND
+                           search_frame = GeomFromWKB(?, 4326))
+                       """
+
+            test_list = self.conn.execute(check_sql, [point.wkb, point.wkb]).fetchone()
+            while sum(test_list):
+                test_list = self.conn.execute(check_sql, [point.wkb, point.wkb]).fetchone()
+                point = Point(point.x + random.random() * shift,
+                              point.y + random.random() * shift)
```
jamiecook, line 97: I would consider extracting this to a method `perturb_point_to_avoid_existing_nodes`
Then you could test it in isolation from the insert to database - and this method would become 

```
if robust:
    point = perturb_point_to_avoid_existing_nodes(point)
```

jamiecook (approved):

aequilibrae/project/zone.py
pedrocamargo: Yeah... It would've been better...

PR #299, opened 2022-03-21
Skimming on Multi-class assignment

pedrocamargo:
Full description in the linked issue


aequilibrae/paths/linear_approximation.py
```diff
@@ -451,13 +458,18 @@ def execute(self):
             else:
                 self.steps_below = 0
 
+            if self.iter < self.max_iter:
```
pedrocamargo, line 461: We only reset the data for the last iteration if we have not exited the loop

aequilibrae/paths/linear_approximation.py
```diff
@@ -426,6 +435,11 @@ def execute(self):
                 self.cores,
             )
 
+            for c in self.traffic_classes:
```
pedrocamargo, line 438: Updates the arrays used for skimming when travel time is skimmed for any of the modes

aequilibrae/paths/traffic_assignment.py
```diff
@@ -576,64 +578,67 @@ def save_skims(self, matrix_name: str, which_ones="final", format="omx") -> None
         if mat_format == "omx" and not has_omx:
             raise ImportError("OpenMatrix is not available on your system")
 
-        file_name = f"{matrix_name}.{mat_format}"
-
         mats = Matrices()
-        export_name = path.join(mats.fldr, file_name)
+        for cls in self.classes:
```
pedrocamargo, line 582: We create a skim matrix for each class in the assignment.  We might do it differently in the future if we have a more capable matrix API

aequilibrae/paths/traffic_class.py
```diff
@@ -41,13 +41,10 @@ def __init__(self, name: str, graph: Graph, matrix: AequilibraeMatrix) -> None:
         self.mode = graph.mode
```
pedrocamargo, line 41: We defer the preparation of the objects so we can set skimming even after creating  traffic classes (needed for the case of the QGIS plugin and overall better usability)

aequilibrae/paths/linear_approximation.py
```diff
@@ -339,7 +341,14 @@ def doWork(self):
...
         for c in self.traffic_classes:
+            # Sizes the temporary objects used for the results
+            c.results.prepare(c.graph, c.matrix)
+            c._aon_results.prepare(c.graph, c.matrix)
+            c.results.reset()
+
+            # Prepares the fixed cost to be used
             if c.fixed_cost_field:
                 # divide fixed cost by volume-dependent prefactor (vot) such that we don't have to do it for
                 # each occurence in the objective funtion. TODO: Need to think about cost skims here, we do
```
janzill, line 354: @pedrocamargo this is a comment I left when we added generalised cost, how does this interact with skimming? When producing travel time skims this is not a problem but can you remind me what options we expose to generate skims?

pedrocamargo: Skims are different arrays in the graph object, where the graph.cost/compact_cost is what is actually used in path computation.

aequilibrae/paths/linear_approximation.py
```diff
@@ -426,6 +435,11 @@ def execute(self):
                 self.cores,
             )
 
+            for c in self.traffic_classes:
+                if self.time_field in c.graph.skim_fields:
+                    k = c.graph.skim_fields.index(self.time_field)
+                    aggregate_link_costs(self.congested_time[:], c.graph.compact_skims[:, k], c.results.crosswalk)
```
pedrocamargo, line 441: This is the line where we put the congested time back into the skim array, if that's being skimmed

image

aequilibrae/paths/linear_approximation.py
janzill: If we do not expose generalised cost for skimming then I don't even need to think about this. Looks good otherwise.

janzill (approved):
given we don't expose generalised cost for skimming this looks good to me

PR #302, opened 2022-03-31
Binaries for different NumPy versions

pedrocamargo:


.github/workflows/compile_windows_binaries.yml
```diff
@@ -0,0 +1,39 @@
...
+        cd aequilibrae/paths
+        python setup_Assignment.py build_ext --inplace
+        cd ..
+        cd ..
+
+    - name: Stores artifacts along with the workflow result
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ matrix.numpy_version }}
+        path: aequilibrae/paths/*.pyd
```
jamiecook, line 38: I like the fail if files missing - but by specifying *.pyd - this will pass as long as ANY pyd are found.
Is there a way to specify explicitly which files are expected?

jamiecook (approved):

Issue #307, opened 2022-04-11
Cannot compute shortest path


Art-Ev:
**Describe the bug**

Traceback (most recent call last):
              File "%FolderProfile%/python/plugins\AequilibraE\paths_procedures\show_shortest_path_dialog.py", line 154, in produces_path
              self.res.compute_path(int(self.path_from.text()), int(self.path_to.text()))
              File "%FolderProfile%\python\plugins\AequilibraE\aequilibrae\aequilibrae\paths\results\path_results.py", line 88, in compute_path
              self.skims[self.graph.all_nodes, :] = self._skimming_array[:-1, :]
             IndexError: index 1066 is out of bounds for axis 0 with size 1066

**To Reproduce**
Use shortest path dialog: 
For example trying to find shortest (distance) path from 387 to 1067 with [St_Varent.zip](https://github.com/AequilibraE/AequilibraE-GUI/files/8462386/St_Varent.zip)

**Desktop (please complete the following information):**
 - OS: windows 10
 - QGIS Version 3.22

pedrocamargo:
@Art-Ev , I will close this issue as it has already been merged into develop and that's the workflow we are using for AequilibraE now

PR #308, opened 2022-05-15
Pedro/issue307

pedrocamargo:
@Art-Ev , this addresses issue #307


tests/aequilibrae/paths/test_pathResults.py
```diff
@@ -174,3 +174,24 @@ def test_update_blocking_trace(self):
...
+        self.project.open(self.st_varent)
+        self.project.network.build_graphs()
+        self.g = self.project.network.graphs["c"]  # type: Graph
+        self.g.set_graph("distance")
+        self.g.set_skimming('distance')
+
+        self.r = PathResults()
+        self.r.prepare(self.g)
+
+        self.r.compute_path(387, 1067)
```
jamiecook, line 196: did you want to assert what the computed path actually is?

jamiecook (approved):

tests/aequilibrae/paths/test_pathResults.py
pedrocamargo: Not really. The only concern is it actually running

PR #309, opened 2022-05-16
Development of version 0.7.4

pedrocamargo:
Fixes

* Path computation when skimming is required and the node with the highest ID is one of the centroids.
* Type for link name was corrected to be TEXT rather than numeric
* Adds free-flow travel time as a default field for the links table
* Remove greek letter fields
* Add field for speed to the link_type table


pedrocamargo:
I made a little change to the Delaunay triangulation code and it is now passing tests. It looks like something related to the project activation setup, but I don't understand why. Something to keep an eye on, @elfjes . Unfortunately a goofed off and committed it directly to develop...

jamiecook (approved):
I feel like this is a just the combo of multiple other already reviewed PRs right?
If you've done a quick check of it all working together in QGIS I'd say LGTM!!!!

PR #318, opened 2022-05-31
Run black on all source files

### LINKED ISSUE #317: Run black on all source files
Hi!

### Preamble
I'm currently preparing a number of issues/PRs for Aequilibrae. With my company we're running a (not yet public) fork of Aequilibrae for which I've recently become the maintainer. You may already have heard of us through our ex-colleague @barisdemirdelen. Now, I would like to look at the possibility to bring (some of) our changes into Aequilibrae proper, which would it easier for us to maintain our fork and gives Aequilibrae some newer/enhanced features/capabilities. Our changes include:
- Support for multiple projects
- Some threading changes (or perhaps customizability/configurability of some threading aspects)
- Some (test) performance changes
- Getting rid of some warnings/deprecations

/end-of-preamble

I've noticed that you are using black (judging from the .pre-commit-config.yaml) for code formatting. This is great! However, not all code seems to be blackened yet. I would like to keep the PRs I'm preparing clean, but right now precommit is blackening every file I'm touching so there's also a bunch of changes that are just code formatting. I would like to suggest to create a single commit/PR that blackens all of the codebase, so any further PRs are not muddied by code format changes. The PR would consist of the following

- update .pre-commit-config.yaml to point to a specific black version
- add a pyproject.toml so that black can easily be run from commandline with the specified line length
- update .gitignore (black automatically ignores everything in .gitignore to ensure it doesn't format files it doesnt need to)
- run black

Now, this commit/PR would muddy git blame, since it would appear that the commiter changed all files. Black proposes a solution for that in the form of an ignore-revision (see https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html). There would be a second commit that adds the first commit's hash to a `.git-blame-ignore-revs` file. This file can then be passed to git blame (and is also automatically passed to Githubs blame server) so that those changed lines would still be attributed to the correct commit.

What do you think of this?

elfjes:
fixes #317

* Updated `.gitignore` with some common lines coming from `github/gitinore/Python.gitignore` and added/modified a few 
lines where I found temporary files being created/modified by Aequilibrae in my local directory
* pointed `pre-commit-config.yaml` to updated refs for both black and flake8
* added a pyproject.toml with the black settings
* blacken everything
  * saved the commit hash in a file and added it to the pr

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,12 +63,34 @@ aequilibrae/paths/build/*
 docs/build/*
 docs/source/_generated/*
 docs/source/_auto_examples/*
+*.cpp 
 
 
 #Virtual environment
 .venv/
 
 # Test files
-tests/*.aem
-tests/*.csv
-tests/*.omx
+tests/**/*.aem
+tests/**/*.csv
+tests/**/*.omx
+
+#### Snippet from https://github.com/github/gitignore/blob/main/Python.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+build/
+*.egg-info/
+*.egg
+
+# Unit test / coverage reports
+.coverage
+.coverage.*
+coverage.xml
+
+#### End snippet
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Migrate code style to Black
+da16bfaffaed76cfdf5e53d96ff155cd4c4a78d5
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,9 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 22.3.0
     hooks:
     - id: black
-      language_version: python3.8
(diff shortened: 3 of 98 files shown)


.gitignore
```diff
@@ -63,12 +63,34 @@ aequilibrae/paths/build/*
 docs/build/*
 docs/source/_generated/*
 docs/source/_auto_examples/*
+*.cpp
```
pedrocamargo, line 66: I am not sure about this.  @janzill ?

janzill: These are cython generated, we switched to c++ when implementing path files and therefore there's a AoN.cpp that should be ignored

pedrocamargo:
This is nice, @elfjes . My only concern is that not everybody that contributes to AequilibraE uses the pre-commit hooks and I don't see changes to the linting process. Is that the case to include something like https://github.com/zheller/flake8-quotes ?

pedrocamargo:
@elfjes , adding that GH action would be perfect.  I am not worried about those tests for now. They usually fail when outside contributors put PRs in because of some secrets stored within GH  (although the failures seem to have happened elsewhere)

elfjes:
I added the `black check` as new step in the linting job, rather than creating a whole new job/action, seemed to be the most performant option. Awaiting workflow approval

PR #323, opened 2022-06-08
Support for multiple projects

elfjes:
Currently, only a single project can be openened by Aequilibrae, stored in the `AEQUILIBRAE_PROJECT_PATH` environment variable. This PR implements support for multiple projects to be loaded at the same time. It does this by binding the classes that deal with project data to a specific project (ie: providing an (optional) project constructor argument). In order to maintain backwards compatibility, it also introduces the concept of an *active project* (see `context.py`). A project activates itself upon opening/loading, but can also be explicitely activated or deactivated. When instantiating a calculation class (such as `TrafficAssignment`) it is possible to give a project as an argument (`TrafficAssignment(project=project)`) or not (`TrafficAssignment()`). In the latter case the currently active project will be used

Having multiple projects open at the same time also means that the table classes such as `Links` can no longer assume there is a single list of `items`. `items` has been moved to the instance instead of a being class variable. It loses its *singleton* behaviour, so care must be taken to only instantiate these classes once per project. These instances are now bound to the project or the project's network (as far as they weren't bound already) and should not be created by other classes

Furthermore this PR consists of the following changes:

- Most database connections are created through `Project.connect()` instead of calling `database_connection` manually
- Changed the exception when a project is not found from `FileExistsError` to `FileNotFoundError`
- Removed some unused imports

## backwards compatibility
This PR is mostly backwards compatible. The active project concept ensures that most classes can still be instantiated without a specific project. There are a few considerations however:

 * some classes/functions now require a project or project path to be given. From the looks of it, these weren't part of the public api, but maybe it's good to have another look. The classes are:
   * BasicTable and derived classes
   * SafeClass and derived classes
   * Matrices
   * FieldEditor
   * Modes
* From the code it was assumed that the `AEQUILIBRAE_PROJECT_PATH` environment variable was only set inside the application when opening a new or existing project and never read from an outside setter. (eg. a user exporting `AEQUILIBRAE_PROJECT_PATH` prior to starting python / importing aequilibrae. In this PR `AEQUILIBRAE_PROJECT_PATH` is no longer used, so if there is any use case for setting AEQUILIBRAE_PROJECT_PATH from outside Aequilibrae, this no longer works (although this functionality could be added again)
* The change from `FileExistsError` to `FileNotFoundError` is a breaking change

What do you guys think? I have not created an Issue for this (can still do that if desired), but just went ahead and created this PR. If you think it needs a different approach, that's also possible, but I thought this is a/the way that integrates best in the current design.


janzill:
First impression is this looks good, thanks! I do not know how this interacts with the QGIS integration, though. I will have a closer look and think about this when I get some time, it's pretty hectic at the moment so this could take a couple of days.

pedrocamargo:
Hey, @jamiecook . Can you take a look at this?  It resembles a little some of the changes you did on polarislib, so you might be the best person to review it.

aequilibrae/distribution/gravity_application.py
```diff
@@ -133,6 +133,7 @@ def __init__(self, **kwargs):
             error (:obj:`str`): Error description
         """
 
+        self.project = project or get_active_project()
```
pedrocamargo, line 136: Beautiful

aequilibrae/project/network/safe_class.py
```diff
@@ -33,3 +34,6 @@ def _save_new_with_geometry(self):
         data.extend([self.geometry.wkb, self.__srid__])
         sql = f'Insert into {self._table} ({",".join(up_keys)}) values({markers})'
         return data, sql
+
+    def conn(self):
```
jamiecook, line 38: I'm a little unsure about calling this method conn - my intuition is that it would return an existing connection, not establish a new one.

aequilibrae/project/zone.py
```diff
@@ -23,7 +22,7 @@ def __init__(self, dataset: dict, zoning):
 
     def delete(self):
         """Removes the zone from the database"""
-        conn = database_connection()
+        conn = self._project.connect()
```
jamiecook, line 25: not a problem with this PR, but I notice here that this db connection isn't closed

tests/aequilibrae/distribution/test_gravityCalibration.py
```diff
@@ -29,10 +30,15 @@
 
 
 class TestGravityCalibration(TestCase):
+    def setUp(self):
+        # GravityCalibration requires an object that has a `parameters` attribute. `Parameters` fits
+        # this requirement, so that we don't need to create a full project
```
jamiecook, line 35: quack

jamiecook (approved):
My thoughts mirror @janzill's - this is really good and something that will help the overall architecture going forward. 

To be clear - this doesn't fix all the problems with global variables but it does condense dramatically the global variables to a single place. 

What I've been trying to do in another similar project is to move all of the global state logic into the QGIS plugin (where it's needed) and keep it out of the core library logic altogether. This is a great step in that direction. 

Things I think need further consideration in a future PR:
- [ ] The level of coupling up and down the chain (networks have links and links have a network, networks have a project and a project has a network)
- [x] Thats it

aequilibrae/project/network/safe_class.py
elfjes: renamed to `connect_db`

tests/aequilibrae/distribution/test_gravityCalibration.py
elfjes: 🦆

elfjes:
Thanks for the reviews, I've processed your comments and fixed a few things that were breaking the pipeline. If there are no additional comments, I think it's ready to be merged

pedrocamargo:
I am sorry for the delay, @elfjes .  Me and my wife got pretty sick for a couple of weeks (still not 100% back) and everything went haywire on my end.

elfjes:
No problem @pedrocamargo. Wish you a speedy recovery. Thanks for merging :)

PR #327, opened 2022-07-12
Support logging for multiple projects, log most warnings instead of raising them

### LINKED ISSUE #324: [Q] warnings.warn vs logger.warning
I've noticed that Aequilibrae sometimes uses `starts_logging.logger.warning` to throw warnings, eg https://github.com/AequilibraE/aequilibrae/blob/c3dfaf0b9f375027018c7a0e099f93734684a259/aequilibrae/paths/graph.py#L368, and sometimes `warnings.warn`, eg https://github.com/AequilibraE/aequilibrae/blob/c3dfaf0b9f375027018c7a0e099f93734684a259/aequilibrae/paths/graph.py#L387

Is there a specific reason for this? I'm asking because I'm looking for a way to disable the warnings raised using `warnings.warn`; we'd like to hide them in our applications. If they could all be given to `logger.warning`, that'd be the simplest approach.

elfjes:
This PR tackles a couple of thing with regards to logging (fixes #324):

* The support for multiple (concurrent) projects means that logging also needs to support concurrent outputs/handlers. This PR implements the following
  * Every project gets its own unique logger with a single filehandler outputting to the project's log file. The logger is stored under `Project().logger`
  * The `aequilibrae.context` module contains a `get_logger` function that returns the currently active project's logger, or the global logger if there is no active project
  * In most situations where a logger is needed, it is easily obtained from the given project
* Whenever possible, `warnings.warn` calls have been replaced with the appropiate `logger.warning` call.
* Examples have been updated
* The global logger has been explicitly name `global_logger` to avoid confusion. The logger is still available under the old name `logger` for backwards compatibility
* Tests are added/updated in pytest style as a proof-of-concept
  * added a `conftest.py` that contains global fixtures
  * a single global fixture has been added that is a factory to create a project (based on the siouxfalls project). Tests (and other fixtures) now only need to use this fixture and call it to create a new project
  * all tests and other fixtures are still bound to a class for constistency with the current unittest style tests

Other fixes
* fixed a few erroneous type hints

noteworthy points/caveats:
* In order to ensure uniqueness, the project logger name now includes the full path to the project's directory. This can be a long name. In order to not clutter the log lines, the logger name has been removed from the log format for both the project loggers as the global loggers. This is a breaking change (for those users that parse the logger output, if any). We can put the name back by hardcoding it into the formatter. However, the logger name ("aequilibrae") does not add much user value, since it's kind of obvious that the logs in `aequilibrae.log` belong to the aequilibrae library/application
* There's a catch-22 in the warning log in `parameters:52`. In order to log this message to the project logger, we need to know whether logging is enabled for this project, which we only know after the Parameters have been setup. I have circumvented it by logging the message to the global logger
* The project loggers do not propagate their log messages to the global logger. This means that there is no single call to update the log output (for example, add a stdout handler to all loggers)

Open questions:
- [x] I'm not sure about the naming `starts_logging` and `StartsLogging`. I don't know the origin of this naming, but do you guys think it's still the right name?


aequilibrae/project/network/connector_creation.py
```diff
@@ -70,7 +70,9 @@ def connector_creation(geo, zone_id: int, srid: int, mode_id: str, network, link
     if len(nodes) == 0:
         raise Exception("We could not find any candidate nodes that satisfied your criteria")
     elif len(nodes) < connectors:
-        warn(f"We have fewer possible nodes than required connectors for zone {zone_id}. Will connect all of them.")
+        logger.warning(
+            f"We have fewer possible nodes than required connectors for zone {zone_id}. Will connect all of them."
```
pedrocamargo, line 74: Did we black the code to 88 characters instead of 120?

pedrocamargo (approved):

pedrocamargo:
I am very happy for you to change the naming. This comes from a day when I had NO idea how logging worked

elfjes:
> I am very happy for you to change the naming. This comes from a day when I had NO idea how logging worked

Sure, will do

pedrocamargo:
> I noticed I'm getting some hard errors (maybe segfaults) during tests sometimes. I initially thought it had something to do with my computer, but now I also see it popping up in CI (see linting / testing 3.7 ubuntu). Is that something you've seen before? I currently have no idea where it comes from

They were more frequent a while back, but I haven't gotten any in some time. If I remember correctly, the errors were related to errors in the Cython code that made it go outside array boundaries, which was hard to detect with the decorators that void that checking (which are in place for performance reasons).

aequilibrae/project/about.py
```diff
@@ -1,11 +1,10 @@
 from os.path import join, dirname, realpath
 import sqlite3
 import string
-from warnings import warn
 import uuid
+from aequilibrae.context import get_logger
```
jamiecook, line 5: Is this method actually used here?

aequilibrae/project/network/link.py
```diff
@@ -107,12 +107,12 @@ def add_mode(self, mode: [str, Mode]):
         mode_id = self.__validate(mode)
 
         if mode_id in self.modes:
-            logger.warning("Mode already active for this link")
+            self._logger.warning("Mode already active for this link")
```
jamiecook, line 110: I don't think that _logger is populated for the link is it? self.project.logger perhaps?

aequilibrae/project/network/osm_downloader.py
```diff
@@ -9,12 +9,13 @@
...
 For the original work, please see https://github.com/gboeing/osmnx
 """
+import logging
 import time
 import re
 import requests
 from .osm_utils.osm_params import http_headers, memory
 from aequilibrae.parameters import Parameters
-from aequilibrae import logger
+from aequilibrae.context import get_logger
```
jamiecook, line 18: again - i don't think that get_logger is used here.

tests/aequilibrae/conftest.py
```diff
@@ -0,0 +1,27 @@
+import uuid
```
jamiecook, line 1:
```suggestion
# The conftest.py file serves as a means of providing fixtures for an entire directory. 
# Fixtures defined in a conftest.py can be used by any test in that package without 
# needing to import them (pytest will automatically discover them).

import uuid
```

jamiecook:
I'm also happy with this PR - a couple of questions/comments/suggestions thrown in. 

The number of CI tests is pretty big at the moment - might be worth rationalising them somewhat if they are getting in the way of actually getting things shipped instead of supporting that.

aequilibrae/project/network/osm_downloader.py
elfjes: it's used on line 39: `self.logger = logger or get_logger()`

tests/aequilibrae/conftest.py
elfjes: thanks :)

elfjes:
Thanks @jamiecook for your review, I've processed your changes.

I'm still looking at the SIGABRT errors, I have no idea where they come from. At my machine they popup sometimes, but only when running the full unit test suite. It seems that in CI it's pretty reproducible for python3.7 but I haven't been able to reproduce it in docker (perhaps stemming from me using python docker image instead of ubuntu).

There also shouldn't be any significant difference between this branch and `develop` with regards to these things. So, no solution yet...

@jamiecook what do you mean when you say the number of CI tests is pretty big? the number of workflows / matrix size of the number of unittests? If it's the number of workflows/matrix size then this could be reduced, although given that they're mostly run in parallel I'm not sure if it would speed up things. If it's about the number of unittests then I disagree, it's the unittests that make it possible to refactor code such as I did with the multiple-projects PR. I do have some ideas to speed up the tests, but that would involve making more use of pytest.

jamiecook:
@elfjes - my comment was more on the number of workflows (currently 17) than the number of unittests. I agree on the utility of unittests. 

However any test-suite that can't be consistently relied upon can sometimes be more of a hinderance than a help for development. Having said that I'm not sure what to do here either - python 3.7 still has a year left before EOL, so should ideally be kept. sigh

elfjes:
> I've added a PR that should get rid of the first two failures from forks - #329

Nice. I do agree that it is important to not have regularly failing pipelines (and accept such a fact). So good stuff on that PR. 

> However any test-suite that can't be consistently relied upon can sometimes be more of a hinderance than a help for development. Having said that I'm not sure what to do here either - python 3.7 still has a year left before EOL, so should ideally be kept. sigh

As a test I added a workflow that performs the tests in a container, which did not crash the tests. So while I still don't know what causes it exactly, it does seem to be some kind of binary incompatibility *somewhere*. My suggestion for now would be to run all (failing) workflows in a container when possible, which seems (n=1) to be mitigating the problem.

We may want to have a separate pr that deals with the workflows more comprehensively. There are some other gains to be made (even if we keep the same number of workflows) for example, having the unit tests depend on the linting is not necessary I'd say, and I think the test setup can also be simplified. This could be done in #329 perhaps?

jamiecook:
#329 is already merged but yes a separate pr to deal with that would be better

what do you mean “ performs tests in a container” I thought all gh actions were done using containers? Or are you saying using separate containers for individual states in a workflow?

jamiecook:
Oh yeah - that makes sense as to why each step is able to use filesystem artifacts from previous steps. I would have expected that VM would itself be a container though. Lets merge this and move discussion to your new branch - create now and mark it [WIP] if you'd like early feedback.

Issue #326, opened 2022-07-11
[Suggestion] use pytest style tests (w/ fixtures etc)


elfjes:
Aequilibrae currently uses pytest as a test runner, but does not make use of any of the (very useful) features of this package, such as fixtures and the assert (rewriting) system, since all tests are still in `unittest` style.  I can see the following benefits of migrating to pytest style testing:

 - using fixtures can further separate setup logic from the test logic
 - fixtures can be shared between different test files to remove code duplication
 - fixtures can have a longer lifespan (scope) than per-test or per-class which means that some expensive fixture functions could only be run once, leading to faster tests. I can see this being especially beneficial for creating an empty project, which currently takes up quite some time
 - the temporary directory system of pytest can help further isolating tests that need to read/write files on disk.
 - opens up further possibilities to create/fill some files at test setup, leading to fewer test data-files that are subsequently modified during testing (yes I'm looking at you test_omx.omx)
 - ability to write `assert something` instead of `self.assert*` is an (arbritrarily) nicer api
 - parametrization to improve test surface / coverage

Would you be willing to migrate to using pytest proper? If so, what would be the prefered way of implementation? My suggestion would be

- new tests use pytest style testing
- whenever existing tests are modified, they are converted to pytest style, gradually building up a fixture ecosystem
- In between, chore PRs are created that focus on converting part of the test code to pytest style
- at some point™ all remaining tests are converted to pytest style tests

In my experience, converting from unittest to pytest can be relatively time consuming, especially if you want to do it correctly with differently scoped fixtures (but the resulting code is much more maintainable and extensible).

pedrocamargo:
This sounds like a good idea for the long term, but I have no extensive experience with more sophisticated use of testing frameworks. We would need to things to get started:

* Guidelines (in the rst documentation) for the minimum structure required for test (something code reviewers would hopefully require before approving a PR
* A test or two converted to the new style.
Are those two things you could take on for a first PR in that direction?

pedrocamargo:
@elfjes , as you guys have been doing quite a bit of work on AequilibraE, you should be aware of this: https://www.linkedin.com/feed/update/urn:li:activity:6978111166732525568/

PR #331, opened 2022-07-26
Refactor tests to pytest

elfjes:
Following #326 I've refactored some more tests to make better use of pytest's capabilities

* added some more global fixtures to request a project:
  * either a clean or a filled project (eg: siouxfalls)
  * with a scope of session or test (to choose between long lived and short lived projects)
* Refactored some tests files. 
* IMHO unit tests should be deterministic (ie, no randomization). There's a place for randomization in testing (ie. property based testing, using `hypothesis` for example) but that requires a structured/systematic approach. Providing a whole bunch of randomized inputs to see if a piece of code succeeds under many different input conditions. Not just a single random value and call it a day
* I've removed `TestFieldEditor.test_check_completeness` as all it did was test the database (running a query and see the result). It didn't test any aequilibrae-code

Take especially note of `test_field_editor.py` which uses some new (ie. not previously introduced here) constructs to remove in-test looping, conditional tests and provide better test isolation:
* test parametrization: using `pytest.mark.parametrize` to introduce new fixtures or override existing ones (first argument) and provide a list of test cases (values of these fixtures)
* fixture parametrization: using `pytest.fixture(params=[...])` to create multiple versions of a fixture. Every test is run with all versions of the fixture (unless the fixture is overridden, eg using `pytest.mark.parametrize`)
* `@pytest.mark.usefixtures` to use a fixture for a test without specifying it as an argument

After these changes running the `tests/aquilibrae/project` directory (on my machine) took 28 seconds, vs 48 seconds previously (~70% faster). There's room for more improvements given that I only refactored a couple of files


pedrocamargo:
This uses stuff that @jamiecook is probably more comfortable with reviewing than me.

tests/aequilibrae/conftest.py
```diff
@@ -12,20 +12,64 @@
...
+        projects.append(project)
+        return project
+
+    yield _create_project
+    for project in projects:
+        project.close()
+
+
+@pytest.fixture
+def create_empty_project(_empty_project, create_project):
```
jamiecook, line 30: this shadows the method above... 

```suggestion
def create_empty_project(_empty_project, create_project_session):
```

tests/aequilibrae/conftest.py
```diff
@@ -12,20 +12,64 @@
...
+    def _create_empty_project(name=None):
+        return create_project(name=name, source_dir=_empty_project)
```
jamiecook, line 32: create_project var shadows above method name.

Maybe for consistency: 

```suggestion
def create_empty_project(_empty_project, create_project_session):
    def _create_empty_project(name=None):
        return create_project_session(name=name, source_dir=_empty_project)
```

jamiecook: urgh - i'm an idiot - never mind - that's HOW fixtures work.

tests/aequilibrae/project/test_create_example.py
```diff
@@ -1,17 +1,20 @@
...
-            pth = join(gettempdir(), f"proj_example_{p}_{uuid4().hex}")
-            proj = create_example(pth, p)
-            parproj = proj.parameters
-            self.assertEqual(parproj.keys(), par.keys(), f"Wrong parameter keys for {p} example")
-            proj.close()
+class TestCreateExample:
+    # We use fixture parametrization here to create the projects so that we can do proper teardown
+    # of the project, even when the test fails
+    # see also: https://docs.pytest.org/en/6.2.x/fixture.html#fixture-parametrize
+    @pytest.fixture(params=["nauru", "sioux_falls"])
```
jamiecook, line 11: I feel this is nice and easy to understand - and it doesn't seem to use any of the helper fixtures from conftest.py right?

tests/aequilibrae/conftest.py
```diff
@@ -12,20 +12,64 @@
...
+    projects = []
+
+    def _create_project(name=None, source_dir=siouxfalls_project):
+        project = create_project_session(name=name, source_dir=source_dir, teardown=False)
+        projects.append(project)
+        return project
+
+    yield _create_project
+    for project in projects:
+        project.close()
```
jamiecook, line 26: something feels a little off here - why are we maintaining two lists of created projects (one here and one in create_project_session). I'm guessing one is persistent across the entire test session and one is only within a given test... but it seems over complicated the way it is setup. Maybe I just don't understand the utility though.

tests/aequilibrae/project/test_field_editor.py
```diff
@@ -1,137 +1,98 @@
...
+from pathlib import Path
+
+import pytest
+from aequilibrae.project.field_editor import FieldEditor
+
+
+class TestFieldEditor:
+    @pytest.fixture(scope="session")
+    def project(self, create_empty_project_session):
+        return create_empty_project_session()
```
jamiecook, line 10: I'm not sure I like that project sometimes has session scope and sometimes test scope - perhaps `project` and `project_session` would be helpful to distinguish. Also - why do they sometimes have different scopes

tests/aequilibrae/project/test_field_editor.py
```diff
@@ -1,137 +1,98 @@
...
+        q2 = curr.fetchone()[0]
+        assert q2 == attribute_count + 1, "Adding element did not work"
+
+        result = curr.execute(
+            f'select "{new_attribute}" from "attributes_documentation" where name_table="{table_name}"'
+        ).fetchone()[0]
+        assert result == new_attribute
+
+    @pytest.mark.parametrize(
+        "table_name, attribute, description",
```
jamiecook, line 76: this is in an interesting use - this table_name isn't consumed in this test, but rather feeds into the "table" fixture which this test consumes. Tricky.

tests/aequilibrae/project/test_create_example.py
elfjes: Nope, it needs to test `create_example` so using the helper fixtures would defeat the purpose ^^

tests/aequilibrae/conftest.py
elfjes: That is mainly for performance reasons. Ideally we would want a fresh project for every test. However, creating a new project takes some time, so when tests only do read actions on a project we can reuse a project through several tests. Looking at `test_field_editor.py` for example, we went from a few large tests that create a project per test, to many smaller tests. If each of those tests would create a fresh project, the test suite would slow down significantly. Instead we create one project for the full test suite, and clean it up (reset it) for those tests that actually modify it.

It is setup in this way because teardown of a fixture only happens once the scope of the fixture has ended, and we want to close the projects as soon as they're no longer needed. We could do all project's closing at session teardown, but then we'd have resources building up (dangling db connection etc) throughout the test session.

Having said that, I can try and refactor these fixtures to make them a little bit less confusing

tests/aequilibrae/project/test_field_editor.py
elfjes: Sure, a rename sounds good. As to the why, see my above comment :)

tests/aequilibrae/project/test_field_editor.py
elfjes: Yes, very weird when you see it the first time. I'll be sure to comment it. Or, if preferable, I can refactor and setup this test some other way

tests/aequilibrae/conftest.py
elfjes: I've refactored these fixtures. Is it better now? or too many nested functions? 😅

tests/aequilibrae/project/test_field_editor.py
jamiecook: Maybe a comment - to help anyone who doesn't know pytest that well. No need to refactor - it's working as the framework was designed

tests/aequilibrae/conftest.py
```diff
@@ -11,21 +11,59 @@
...
-    for project in projects:
-        project.close()
+def create_empty_project_session(_empty_project, create_project_session):
+    def _create_empty_project(name=None):
+        return create_project_session(name=name, source_dir=_empty_project)
+
+    return _create_empty_project
+
+
+@pytest.fixture(scope="session")
```
jamiecook, line 59:
```suggestion
# This fixture creates a default empty structure on disk that can be used as the source folder for creating temporary empty projects
@pytest.fixture(scope="session")
```

tests/aequilibrae/project/test_field_editor.py
```diff
@@ -1,137 +1,101 @@
...
+from pathlib import Path
+
+import pytest
+from aequilibrae.project.field_editor import FieldEditor
+
+
+class TestFieldEditor:
+    @pytest.fixture(scope="session")
+    def project_session(self, create_empty_project_session):
+        return create_empty_project_session()
```
jamiecook, line 10: Would it make sense to now move this next to the `project` fixture in the conftest.py?

tests/aequilibrae/project/test_field_editor.py
```diff
@@ -1,137 +1,101 @@
...
+    def database_backup(self, project_session):
+        return Path(project_session.project_base_path).joinpath("project_database.sqlite").read_bytes()
+
+    @pytest.fixture
+    def cleanup_database(self, project_session, database_backup):
+        """An optional fixture that can be used to revert the project's database in case the test
+        modifies it. This adds some extra overhead to the test (teardown)
+        """
+        yield
+        Path(project_session.project_base_path).joinpath("project_database.sqlite").write_bytes(database_backup)
```
jamiecook, line 22: Same with these... they seem quite generic and useful in other tests.

tests/aequilibrae/project/test_field_editor.py
jamiecook: comment looks perfect.

tests/aequilibrae/project/test_mulitple_projects.py
```diff
@@ -1,52 +1,29 @@
...
 
 
-class TestMultipleProjects(TestCase):
-    @classmethod
-    def setUpClass(cls):
-        cls.proj = cls.new_project()
+class TestMultipleProjects:
+    @pytest.fixture(scope="session")
+    def project(self, create_empty_project_session):
+        return create_empty_project_session()
```
jamiecook, line 8: use the renamed one from conftest.py

tests/aequilibrae/conftest.py
jamiecook: Looks much better - I think i was unclear in my comment - I understood why you would want different scopes, but the previous impl called from one to the other and had that janky flag to say whether things added should be cleaned up.... this is much better.

tests/aequilibrae/project/test_field_editor.py
elfjes: I'm a little bit hesitant to do this. Having a session scoped project inside a class is one thing, having a global sessions scoped project is anothe. Currently, if a test behaves poorly and doesn't clean up after itself, it can only affects the other tests inside the class. In case of a global fixture that's used throughout the whole test suite there's an increased risk of tests affecting other tests in unpredictable ways.

Also, in the `TestFieldEditor` class we know the only part of the Project that is mutated is the sqlite database, so we also only need to restore that. For the general case, we don't know that, so we also cannot make the assumption that restoring the database is sufficient. So we'd need to restore the whole project, which is a significantly more expensive operation (and is effectively what we do with `create_project_session` already)

jamiecook: Hmmm - my understanding of scope is that if it's defined with session scope - it won't matter where it's defined - there will only be one of them for the entire test suite. Or are you relying here on the fact that each of the fixtures is namespaced to the class that is invoking it? I think that makes sense - although it could therefore be tagged as "module" level scope.

tests/aequilibrae/conftest.py
jamiecook: does this comment make sense @elfjes ?

tests/aequilibrae/project/test_field_editor.py
jamiecook: and I get what you are saying here - these cleanup scripts are "subsets of a full wipe and rebuild" that make sense if we know we are just testing "field editor".

tests/aequilibrae/project/test_field_editor.py
jamiecook: If what I said lines up with your understanding - i'm happy to resolve this conversation and approve.

jamiecook (approved):

tests/aequilibrae/conftest.py
elfjes: yes it does, I've comitted it :)

elfjes:
I've process the latest comments. If there's no other remarks, I think this can be merged :)

PR #341, opened 2022-10-13
Updates actions

pedrocamargo:
These actions need updating because support for older versions is about to be discontinued


pedrocamargo:
Some stuff started breaking, and it looks like it is something to do with behavior changes in Numpy, so another PR needs to come before this one

pedrocamargo:
@djfrancesco , We can re-run the tests after #342 is merged, as then they should pass

jamiecook (approved):

PR #342, opened 2022-10-13
Pedro/fixes numpy changes

pedrocamargo:
Some tests had broken because of changes in behavior for NumPy

### DIFF
--- a/aequilibrae/distribution/gravity_calibration.py
+++ b/aequilibrae/distribution/gravity_calibration.py
@@ -218,15 +218,15 @@ class GravityCalibration:
         mats = [(self.matrix, "Observed matrix"), (self.impedance, "Impedance matrix")]
         for matrix, title in mats:
             if matrix.matrix_view is None:
-                raise ValueError(title + " needs to be set for computation")
+                raise ValueError("{title} needs to be set for computation")
             else:
                 if len(matrix.matrix_view.shape[:]) > 2:
-                    raise ValueError(title + "' computational view needs to be set for a single matrix core")
+                    raise ValueError("{title}' computational view needs to be set for a single matrix core")
 
             if np.nansum(matrix.matrix_view.data) == 0:
-                raise ValueError(title + "has only zero values")
+                raise ValueError(f"{title} has only zero values")
             if np.nanmin(matrix.matrix_view.data) < 0:
-                raise ValueError(title + "has negative values")
+                raise ValueError(f"{title} has negative values")
 
         # Augment parameters if we happen to have only passed one
         default_parameters = self.__get_parameters()
--- a/aequilibrae/distribution/gravity_calibration.py
+++ b/aequilibrae/distribution/gravity_calibration.py
@@ -218,11 +218,9 @@ class GravityCalibration:
         mats = [(self.matrix, "Observed matrix"), (self.impedance, "Impedance matrix")]
         for matrix, title in mats:
             if matrix.matrix_view is None:
-                raise ValueError("{title} needs to be set for computation")
-            else:
-                if len(matrix.matrix_view.shape[:]) > 2:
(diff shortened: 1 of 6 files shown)


pedrocamargo:
@djfrancesco I also formatted a few strings since I was at it...

aequilibrae/distribution/gravity_calibration.py
```diff
@@ -218,15 +218,15 @@ def __check_inputs(self):
         mats = [(self.matrix, "Observed matrix"), (self.impedance, "Impedance matrix")]
         for matrix, title in mats:
             if matrix.matrix_view is None:
-                raise ValueError(title + " needs to be set for computation")
+                raise ValueError("{title} needs to be set for computation")
             else:
                 if len(matrix.matrix_view.shape[:]) > 2:
-                    raise ValueError(title + "' computational view needs to be set for a single matrix core")
+                    raise ValueError("{title}' computational view needs to be set for a single matrix core")
```
djfrancesco, line 224: f prefix seems to be also missing for f-string? And maybe it is intentional but there is an isolated quote ' ?

aequilibrae/distribution/gravity_calibration.py
```diff
@@ -218,15 +218,15 @@ def __check_inputs(self):
         mats = [(self.matrix, "Observed matrix"), (self.impedance, "Impedance matrix")]
         for matrix, title in mats:
             if matrix.matrix_view is None:
-                raise ValueError(title + " needs to be set for computation")
+                raise ValueError("{title} needs to be set for computation")
             else:
                 if len(matrix.matrix_view.shape[:]) > 2:
```
djfrancesco, line 223: Maybe we could use something like `matrix.matrix_view.ndim > 2` ?

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -1021,8 +1021,7 @@ def nan_to_num(self):
             raise NotImplementedError("This operation does not make sense for OMX matrices")
 
         if np.issubdtype(self.dtype, np.floating) and self.matrix_view is not None:
```
djfrancesco, line 1023: Sometimes it helps for readability to put parenthesis in conditions : 

```python
if np.issubdtype(self.dtype, np.floating) and (self.matrix_view is not None):
```

djfrancesco (commented):
Hi Pedro, I am not familiar with the source code yet, so I do not understand precisely what is done here. I still left a few comments and hope this helps

aequilibrae/distribution/gravity_calibration.py
jamiecook: good catch - fixed.

aequilibrae/matrix/aequilibrae_matrix.py
jamiecook: I believe that tools like black will remove unnecessary parentheses... I don't particularly find either better or worse.

aequilibrae/distribution/gravity_calibration.py
```diff
@@ -218,15 +218,15 @@ def __check_inputs(self):
         mats = [(self.matrix, "Observed matrix"), (self.impedance, "Impedance matrix")]
         for matrix, title in mats:
             if matrix.matrix_view is None:
-                raise ValueError(title + " needs to be set for computation")
+                raise ValueError("{title} needs to be set for computation")
```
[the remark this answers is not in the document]
jamiecook: fixed

aequilibrae/distribution/gravity_calibration.py
jamiecook: agreed

jamiecook (commented):
address issues from code-review

jamiecook (approved):

PR #345, opened 2022-10-17
Fixes issue #344

### LINKED ISSUE #344: Gravity estimation/calibration too slow for large matrices
The gravity application/calibration code is too slow, especially on machines with slower disks.  The issue is most likely caused by memory-mapping temporary matrices.  Since this memory-mapping uses memory anyways, switching to memory-only matrices is the way to go.

pedrocamargo:
This is something that becomes an issue for large models (eg 15k zones), which I have recently faced.

Closes #344

### DIFF
--- a/aequilibrae/distribution/gravity_application.py
+++ b/aequilibrae/distribution/gravity_application.py
@@ -1,18 +1,19 @@
-import sys
-import numpy as np
-import os
-import tempfile
-from time import perf_counter
 import glob
-import logging
-from uuid import uuid4
-from datetime import datetime
 import importlib.util as iutil
+import logging
+import os
+import sys
+import tempfile
+from datetime import datetime
+from time import perf_counter
+from uuid import uuid4
+
+import numpy as np
+
 from aequilibrae.context import get_active_project
 from aequilibrae.distribution.ipf import Ipf
 from aequilibrae.distribution.synthetic_gravity_model import SyntheticGravityModel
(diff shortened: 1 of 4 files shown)


djfrancesco:
From what I understand, you load the full matrix in memory in order to accelerate the access to data, because of  np.mmemaps is kind of lagging with large matrices. Maybe some options of memmaps can be improved, for example regarding the column-wise or raw-wise order? I will read more abour memmaps' efficiency. 
That's another story but did you also ever consider storing the matrix in SQlite ? Or in a Parquet file ?

djfrancesco (approved):

pedrocamargo:
@djfrancesco , That's exactly the idea.  In this case I am just adding the capability of having a memory-backed array, which is a basic feature.

On the storage side, the format is fixed because it is tied in the rest of the code (assignment in Cython, most importantly).  SQLite is too slow, and Parquet has been out of question because it is not available on QGIS for Windows.

pedrocamargo:
Ideally we would have the AequilibraEMatrix be just an internal format (e.g. to feed into assignment) and have all actual storage done in OMX

djfrancesco:
Pedro, I know that this issue is closed but I wanted to make sure I understand it. Did the slow behavior with large memmap matrices also happen with IPF ? I am trying to reproduce it with a simple example.

pedrocamargo:
Exactly, but that only becomes a problem with very large matrices and slower hard drives (i.e. low IOPS in cloud computing

PR #347, opened 2022-10-17
bug fix

pedrocamargo:
A small but important bug in the distribution code


djfrancesco (approved):
I guess I need to read the AequilibraeMatrix class carefully, in order to fully understand these changes!

PR #351, opened 2022-10-24
Version 0.7.5

pedrocamargo:
Point release for the 0.7.5 version


jamiecook:
Why are 1 years worth of commits coming in? Is it really Nov 2021 that 0.7.4 was released?

pedrocamargo:
Yeah... One thing we need to start doing is to put releases on a schedule... Twice a year is easy, but 3 times per year might be perfectly reasonable

pedrocamargo:
@TTalex , @Art-Ev , @jamiecook   Is there anything you believe we should do before we release this version (last one before the change in model structure)?

djfrancesco:
@pedrocamargo right now I don't think of anything more to add to this release (with lots of changes!). I am still in the process of exploring the source code. I looked at issue #344 and did not find yet a nice way to use out-of core matrices for calibration/estimation, regarding both CPU and memory. 
I agree with you that 2 or more releases per year would be great, even if they are kind of small!

jamiecook (approved):
Lots here but it all seems to be previously reviewed and merged. 

Lets get this out the door.

Art-Ev:
> @TTalex , @Art-Ev , @jamiecook, @djfrancesco
> 
> Is there anything you believe we should do before we release this version (The last one before the change in model structure)?

I think we are good, no more "obvious" blocking issues : AequilibraE can be used almost easily in projects on "simple" cases, with no real issues (no more centroid or assignment issues) and creation of needed skims is great ! Next changes will be about new "structure" and new functions

PR #356, opened 2022-11-07
Makes PyArrow an optional dependency

pedrocamargo:


setup.py
```diff
@@ -1,25 +1,33 @@
...
 
-here = os.path.dirname(os.path.realpath(__file__))
-whole_path = os.path.join(here, "aequilibrae/paths", "AoN.pyx")
-ext_module = Extension(
-    "aequilibrae.paths.AoN", [whole_path], include_dirs=[np.get_include(), pa.get_include()], language="c++"
-)
+    include_dirs.append(pa.get_include())
+
+whole_path = join(dirname(os.path.realpath(__file__)), "aequilibrae/paths", "AoN.pyx")
+ext_module = Extension("aequilibrae.paths.AoN", [whole_path], include_dirs=include_dirs, language="c++")
```
jamiecook, line 22: What happens if you build AoN without pyarrow in teh includes list?

jamiecook (approved):

setup.py
pedrocamargo: It crashes (in a rather ugly way) when we try to use with it

djfrancesco (approved):

setup.py
```diff
@@ -1,25 +1,33 @@
...
 
 # this is for building pyarrow on platforms w/o wheel, like our one of our macos/python combos
 if "WINDOWS" not in platform.platform().upper():
     ext_module.extra_compile_args.append("-std=c++11")
```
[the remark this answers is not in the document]
jamiecook: it is mentioned in the requirements for building arrow from source - but seems to work okay when only using the headers for pyarrow (or at least for the subset of bits we are using here)

PR #369, opened 2022-12-15
Transpose shared matrices to better cache locality

Jake-Moss:
By bringing the thread id to the 0-axis we are able to make the thread local slices continuous in memory. This should improve cache locality for better performance in tight loops.

Benchmarked on the Arkansas state wide network and saw ~10% improvement in overall skimming runtime.


pedrocamargo:
Documentation is failing due to change in behavior in Shapely. I will update the documentation on a separate PR

jamiecook (approved):

pedrocamargo:
This can be merged after merging #370 and re-running tests

PR #371, opened 2023-01-06
Make it obvious that this is the MIT licence

jamiecook:


pedrocamargo (approved):

pedrocamargo:
@jamiecook , There is some upstream issue with the windows-2022 runner (which just became windows-latest), so tests won't pass. I think we need to set the test runner to windows-2019 and put an issue to rever this in the future.

PR #374, opened 2023-01-10
Replace the internal Fibonacci heap with a 4-ary priority queue

Jake-Moss:
This PR provides up to a 3.5x performance increase on the various models tested.
It also changes the default behaviour of `memory_only` mode to default to `True`, enables the `NPY_NO_DEPRECATED_API` and `NPY_1_7_API_VERSION` macros for Unix builds, and restores the `language_level` to `3` for Cython.

The heap implementation were written by @djfrancesco and has the MIT license preserved as a header in the `pq_4ary_heap.pyx` file.


aequilibrae/paths/basic_path_finding.pyx
```diff
@@ -14,7 +14,7 @@
  Licence:     See LICENSE.TXT
  -----------------------------------------------------------------------------------------------------------
 Original Algorithm for Shortest path (Dijkstra with a Fibonacci heap) was written by Jake Vanderplas <vanderplas@astro.washington.edu> under license: BSD, (C) 2012
```
jamiecook, line 16: probably need to change this.

Jake-Moss: Good catch, I've updated the other banner as well

djfrancesco (approved):

djfrancesco (approved):

pedrocamargo (approved):

pedrocamargo:
For reference. Image created by @chrisobi02 
MicrosoftTeams-image

pedrocamargo:
Congrats to @Jake-Moss , @chrisobi02 & @djfrancesco for the amazing work that went into this PR!!!

Issue #312, opened 2022-05-20
Testing not working in MacOS


pedrocamargo:
Testing on MacOC could not be set because it is apparently using clang instead of gcc.

Some work on the CI may be needed to fix that 

 https://github.com/AequilibraE/aequilibrae/pull/292

pedrocamargo:
Should we include this in CI since you have figured it out, @jamiecook ?

PR #381, opened 2023-02-05
Fixes issue #312

pedrocamargo:


pedrocamargo:
closes #312

pedrocamargo:
@jamiecook , I tried to include MacOS testing again, but it is segfaulting, which may actually be a problem. Any insights?

PR #385, opened 2023-03-27
updates default endpoint

pedrocamargo:
The old endpoint we were using was deprecated.

https://wiki.openstreetmap.org/wiki/Overpass_API


jamiecook (approved):

jamiecook:
I also have this in py311 branch.

PR #390, opened 2023-03-31
Include requirements.txt when packaging

Jake-Moss:


Jake-Moss:
I've tested with `testpypi` and `manylinux` and this seems to work

pedrocamargo (approved):

PR #398, opened 2023-04-06
Critical test preventing the use of the software altogether

pedrocamargo:


aequilibrae/paths/__init__.py
```diff
@@ -17,7 +17,6 @@
         one_to_all,
         skimming_single_origin,
         path_computation,
-        VERSION_COMPILED,
```
pedrocamargo, line 20: This obviously fails, preventing us from importing most things

r-akemii (approved):

PR #399, opened 2023-04-14
Bring graph compression down to Cython

Jake-Moss:
Currently this yields an approximate ~~2.7x~~ 3.9x increase in performance (10s down to ~~3.7~~ 2.55s on a 5600x) when using the Long An model. More testing and proper benchmarking is to come.

I've looked into parallelising this code but if done naively a thread local `slink` variable, acting as a counter, would provide random access to the `compression_.*` arrays. I'm not exactly sure how to move forward with parallelising it without some more invasive changes.

Some clean up commits are to come regarding the setting of members which is currently done at the end of the cythonised function.


Jake-Moss:
Locally I have two tests failing, due to this conversion  to `np.int8` on line `296`, as direction is only ever in `{-1, 0, 1}` should we look to convert the current usage of `long` to  a smaller data type?
```
FAILED tests/aequilibrae/transit/test_lib_gtfs.py::test_map_match - ValueError: Buffer dtype mismatch, expected 'long' but got 'signed char'
FAILED tests/aequilibrae/transit/test_pattern.py::test_map_match - ValueError: Buffer dtype mismatch, expected 'long' but got 'signed char'
```
https://github.com/AequilibraE/aequilibrae/blob/5dc624642f582243b402a474fc914b5a0319aff6/aequilibrae/transit/map_matching_graph.py#L293-L305

pedrocamargo:
It would be interesting to see benchmarking of the overall graph creating time as well, as that is that what matters in the end (even though the proportional gain would look the same given the preponderance of the compression step).

On the parallelization note, I do have a few ideas on how to go about it, but it might be too big of a lift to do it now. Let's discuss it and write them down for later, though.

On types, I am happy to bring the direction to a smaller type early on.

Jake-Moss:
For reference, here's a benchmark of the current and improved graph compression 
image
This was done using https://github.com/outerl/aequilibrae_performance_tests with
```sh
./benchmark.py --without-project-init -m ./models -p sioux_falls chicago_sketch LongAn Arkansas australia -l aequilibrae_graph_creation -r 3 -i 1 -c 0 --filename graph_creation_after --details after
```

pedrocamargo (approved):

PR #419, opened 2023-05-31
Parallelising PuT assignment

Jake-Moss:
An attempt at parallelising the PuT assignment using Cython.

TODO:
- [x] data type mismatches in buffers (should be easy fix)
- [x] Use of `vector` in `compute_SF_in` type signature. ~Need to find a way to construct a memoryview object from a pointer and size without gil~ This has been side stepped by just using pointers the whole time
- [x] Need to accumulate results into single array, currently they are being discarded. Could construct `threads x number of edges` matrix and have `_SF_in_second_pass` optionally accumulate into it, then reduce over that single threaded at the end
- [x] ~Replace initialisation in `compute_SF_in` with `memset`s~ Compiler should be able to optimise the `0.0` sets (look at .cpp output). `memset` isn't applicable to multi-byte values so it cannot be used with `DATATYPE_INF`
- [x] Remove `maybe` prefix (just used to denote old vs new when debugging)

### DIFF
--- a/aequilibrae/paths/hyperpath.pyx
+++ b/aequilibrae/paths/hyperpath.pyx
@@ -11,7 +11,8 @@ import numpy as np
 cimport numpy as cnp
 
 from libc.stdlib cimport malloc, free
- 
+from libcpp.vector cimport vector
+
 
 ctypedef cnp.float64_t DATATYPE_t
 DATATYPE_PY = np.float64
@@ -100,7 +101,7 @@ cdef void _coo_tocsc_uint32(
 @cython.wraparound(False)
 @cython.embedsignature(False)
 @cython.initializedcheck(False)
-cdef void argsort(DATATYPE_t[::1] data, cnp.uint32_t[:] order) nogil:
+cdef void argsort(cnp.float64_t[::1] data, cnp.uint32_t[:] order) nogil:
     """
     Wrapper of the C function qsort
     source: https://github.com/jcrudy/cython-argsort/tree/master/cyargsort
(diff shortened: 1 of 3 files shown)


Jake-Moss:
This is now parallelised along a single thread and a single thread only. Need to move allocation of thread local arrays into parallel() section.

Currently all threads would slam the same scratch space. Unfortunately in the parallel section, the gil must be released
so we cannot allocate using numpy, instead I've elected to manually manage some pointers for now.

We could allocate large matrices using numpy and take slices but we don't need the scratch space to live any longer than the assignment. Though this is more of an option for the result accumulation (which currently just acquires the gil)

djfrancesco (commented):
Did you try to run the Spiess & Florian test case (aequilibrae/tests/test_hyperpath.py) ? This has only one destination, so it is not really interesting regarding the parallel case. I will try to create a small test case with several destinations, to test the parallel version when you are done.

Jake-Moss:
Yes it pass the `assign` test in its current state, but not the `run` (it is currently commented out as I've changed the interface)

Jake-Moss:
@djfrancesco This is now functional on Linux system. The CI seems to have issues with openmp and Windows but I'll be able to take a look at those on Friday. Otherwise this pr is largely complete.

I'll grab some performance metrics on Friday but just as a quick estimate on my Ryzen 5 5600X using the Lyon notebook, 1 thread takes ~5.6s, with 2 threads ~2.7s, and with 6 threads ~900ms

Jake-Moss:
With a quick benchmark on the Lyon model
image

Before this PR the model took ~5.3s to run, single threaded with this PR it still takes ~5.3s indicating no noticeable performance overhead

djfrancesco:
image

PR #425, opened 2023-06-28
fix matrix indices for memory-only matrices

### LINKED ISSUE #424: Invalid dtype for AequilibraeMatrix index for memory_only matrix
First of all, I like the new(ish) support for memory only matrices, and the performance benefits it offers. Edit: is it relatively new, or did just the default change recently?

However, I did run into an issue. When creating the indices for the AequilibraeMatrix, the numpy array has a `float` dtype, which means it can't actually be used for indexing: 
```
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

So I created a fix, PR coming up.

elfjes:
* fixes #424
* rewrites some tests to pytest
* specify some exception types

### DIFF
--- a/aequilibrae/matrix/aequilibrae_matrix.py
+++ b/aequilibrae/matrix/aequilibrae_matrix.py
@@ -601,7 +601,7 @@ class AequilibraeMatrix(object):
 
         # Index
         if self.__memory_only:
-            self.indices = np.zeros((self.zones, self.num_indices))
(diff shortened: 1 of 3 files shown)


pedrocamargo:
I like what I see here, but I'll ask @jamiecook to take a look here too...

pedrocamargo (approved):

PR #427, opened 2023-06-29
License update

pedrocamargo:


Art-Ev (approved):
Good addition ! Side question : currently aequilibrae uses a non-copyleft licence, does anybody in the "active" community have a plan for some kind of derived-product commercialisation ? A Copyleft licence is more restrictive but can sometimes also allow more contributions to the project...

pedrocamargo:
I have chosen the MIT license precisely to be a little more business-friendly. A few private companies do use AequilibraE in their own products, but some choose to contribute back and we have a good example here: https://github.com/AequilibraE/aequilibrae/pull/425

PR #435, opened 2023-07-19
Matrix exports and deprecation warning

pedrocamargo:
The matrix exporter was not working properly, and we were getting a lot of deprecation warnings.

### DIFF
--- a/aequilibrae/matrix/aequilibrae_matrix.py
+++ b/aequilibrae/matrix/aequilibrae_matrix.py
@@ -3,6 +3,7 @@ import importlib.util as iutil
 import os
 import tempfile
 import uuid
+import warnings
 from functools import reduce
 from typing import List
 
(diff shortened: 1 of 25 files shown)


aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -677,6 +678,8 @@ def __flush(self, obj_to_flush: np.memmap):
 
     def __getattr__(self, mat_name: str):
         if mat_name in object.__dict__:
+            if mat_name == "matrix" and self.__omx:
+                warnings.warn("You can't access OMX matrix cores like that")
```
Jake-Moss, line 682: Should there be an early exit here? If the use can't access the matrix like that should it still return `self.__dict__[mat_name]`? Also warnings are only raised once per source file location so repeatedly hitting this will be silent

Jake-Moss (approved):
LGTM. Theres a few more on related deprecation warnings when running locally but I can tackle those at some point.

PR #440, opened 2023-08-18
Add implicit noexcept from Cython<3.0.0

Jake-Moss:
With the release of Cython 3.0.0 there are a few changes of note to us. Particularly the performance impact of the removal of the implicit noexcept.

Now all cdefs allow exceptions by default meaning every cdef must require the gil at the end of the function regardless of whether it was nogil or not. Adding the noexcept clause reverts to the old behaviour.

There is a compiler directive to result this old behaviour but better be be explicit when the solution is one regex away: `^(cp?def(?:.|\n).*?)(nogil|):$`

https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept

### DIFF
--- a/aequilibrae/distribution/ipf_core.pyx
+++ b/aequilibrae/distribution/ipf_core.pyx
@@ -56,7 +56,7 @@ cdef  _fratar(double[:, :] flows,
                    double[:] attr_factor,
                    int max_iter,
                    double toler,
-                   int cpus):
+                   int cpus) noexcept:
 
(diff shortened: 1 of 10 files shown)


pedrocamargo:
Do we need to make it explicit that we require cython >=3.0 now?

Jake-Moss:
No this is a backwards compatible change

pedrocamargo (approved):

PR #448, opened 2023-09-29
Connector project line string matching

Jake-Moss:
This PR adds project matching for the line strings of the access and egress connectors.

It takes the start and end of the edge, finds the closest nodes within the existing project then computes the shortest path between those nodes and constructs the relevant line string. This can be used to provide more accurate travel time computation and visually appealing results.

TODO:
- [x] Use the more accurate travel time from `line.length / self.walking_speed * self.(access|egress)_time_factor`
- [x] Test on a real network. This code is not tested in its entirety, only on a toy example separate from the rest of the graph building.
- [x] The code currently assumes the project graph has already been initialised correctly and distance is a valid metric. This might end to be improved UX wise.

Note: this does not find the shortest path between the edge ends as visible in the image below.
connectors

### DIFF
--- a/aequilibrae/transit/transit_graph_builder.py
+++ b/aequilibrae/transit/transit_graph_builder.py
@@ -1120,7 +1120,6 @@ class SF_graph_builder:
 
             # Loop over connect edges, query for the clostest nodes in the project and create the relevant line string
             lines = []
-            missing = []
             for row in self.edges[connector_rows].itertuples():
                 # row.a_node - 1 because the node_ids are the index + 1
                 start = shapely.ops.transform(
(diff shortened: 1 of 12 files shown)


djfrancesco (approved):

Jake-Moss:
I'm now happy with how the project matching looks. Heres the results on the Lyon network.
image

image

The intermediate line strings are now constructed from the existing line geometry giving a higher more detailed result. The terminal strings are still straights lines.

For ODs and stops where there doesn't exist a path between them in the project network straights lines are used. This doesn't occur often but looks a little weird when they are clearly connected but the closest node to either end is disconnected.
image
Here the closest node to the origin in the bottom left is disconnected, but another closer one is connected.

aequilibrae/transit/transit_graph_builder.py
```diff
@@ -1058,18 +1061,103 @@ def create_edges(self):
...
+            self.edges["geometry"] = [
+                shapely.LineString(
+                    (
+                        shapely.from_wkb(self.vertices.at[row.a_node - 1, "geometry"]),
+                        shapely.from_wkb(self.vertices.at[row.b_node - 1, "geometry"]),
+                    )
+                ).wkb
+                for row in self.edges.itertuples()
+            ]
+        elif method == "connector project match":
```
jamiecook, line 1078: probably makes sense to extract a method rather than having 100 lines in a else statement

djfrancesco (approved):

PR #456, opened 2023-10-12
Fix documentation bugs on latest

r-akemii:
Modifies the requirement-docs file to use pydata theme until version 0.13.3

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -4,8 +4,8 @@ PyQt5>=5.10.1
 pyshp
 cython
 enum34>=1.1.6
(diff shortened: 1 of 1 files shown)


r-akemii:
@pedrocamargo I've made a simple change in the documentation requirements that should fix the problem in the website layout. When running the files on my local machine, I noticed that the Pydata Theme version was different from the one currently in use on the website and the layout wasn't broken. I did a bit of research and other Python libraries that also use this theme made the same change in their documentation and nothing appears to be broken at their websites. Thus it is probable that the layout will be like the one we had in version 0.9.3.

pedrocamargo (approved):

PR #461, opened 2023-10-31
Adds zones and connectors to Coquimbo.

r-akemii:
Adds traffic zones and centroid connectors to Coquimbo.zip in reference files.

Added one connector per zone for cars and transit, following the same guidelines as in the "Creating zones" example.


Jake-Moss:
Wonderful thank you. I'll take a look at the details and try to use them on Friday

Jake-Moss (approved):
Thanks for this, data was exactly what I needed. 

Just for reference,
image

PR #463, opened 2023-11-03
Better default import from OSM and description of bicycle & bus data in network

Art-Ev:
Parameters file is a bit hidden/tricky to use for qaequilibrae users, this should help them for a better default OSM import which keep data for bus and bicycle
Have keep the same structure as OSM but this result in 3 fields for each one (3 for bicycle and 3 for bus).

Maybe we can think of a way to achieve only one field for bicycle and one for bus ? Might not so easy because of the number of combinations available in OSM, example : https://wiki.openstreetmap.org/wiki/Key:cycleway


Art-Ev:
Don't know why there is 2 commits in this PR, supposed to contain only one

jamiecook:
@Art-Ev - Looks like you maybe merged develop onto your branch?
I've created another PR based on just your commit here: https://github.com/AequilibraE/aequilibrae/pull/464 which i will now review and merge.

PR #464, opened 2023-11-05
Better import from OSM and description of bicycle & bus data in network

jamiecook:
Parameters file is a bit hidden/tricky to use for qaequilibrae users, this should help them for a better default OSM import which keep data for bus and bicycle Have keep the same structure as OSM but this result in 3 fields for each one (3 for bicycle and 3 for bus).

Maybe we can think of a way to achieve only one field for bicycle and one for bus ? Might not so easy because of the number of combinations available in OSM, example : https://wiki.openstreetmap.org/wiki/Key:cycleway

### DIFF
--- a/aequilibrae/parameters.yml
+++ b/aequilibrae/parameters.yml
@@ -42,6 +42,30 @@ network:
           description: name
           osm_source: name
           type: text
+      - cycleway:
+          description: cycleway, both way
+          osm_source: cycleway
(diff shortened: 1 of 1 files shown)


jamiecook:
@pedrocamargo - this seems fine to me, but I can't approve as I created it for @Art-Ev.

janzill (approved):

PR #466, opened 2023-11-14
Early exit Dijsktra's with path reuse

Jake-Moss:
Add a disabled by default option to exit once a destination is found in `path_finding` to  `compute_path` and `update_trace`.

If a partial shortest path tree is computed, `update_trace` will check if the new destination has already been found within that tree, if it isn't it will recompute the necessary tree. It does this by marking the un-scanned nodes as unreachable upon exit, then if the new destination is unreachable and the tree is partial, the tree must be recomputed. Otherwise we are free to reuse the existing tree.

I've added some tests that check the partial trees of the triangle example against what I worked out by hand.

I've also tested this as a default *on* option and all tests still pass.

### DIFF
--- a/aequilibrae/paths/AoN.pyx
+++ b/aequilibrae/paths/AoN.pyx
@@ -114,6 +114,7 @@ def one_to_all(origin, matrix, graph, result, aux_result, curr_thread):
                                     original_b_nodes_view)
 
         w = path_finding(origin_index,
+                         -1,  # destination index to disable early exit
                          g_view,
                          b_nodes_view,
                          graph_fs_view,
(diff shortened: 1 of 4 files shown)


pedrocamargo (approved):

pedrocamargo:
It looked correct to me, but this is an important change, so can you check as well, @jamiecook ?

aequilibrae/paths/results/path_results.py
```diff
@@ -140,19 +145,24 @@ def reset(self) -> None:
         else:
             raise ValueError("Exception: Path results object was not yet prepared/initialized")
 
-    def update_trace(self, destination: int) -> None:
+    def update_trace(self, destination: int, early_exit: bool = False) -> None:
```
pedrocamargo, line 148: I would not add the early_exit here and just use the same in the path_computation.  This is certainly an extra feature, but I am not sure it would ever be used and it may just create confusion.

Jake-Moss: I've removed this arg and added a private variable to track the status of the tree instead. If they want to affect the behaviour of the `compute_path` call within `update_trace`, the public var `PathResults.early_exit` can be set .

pedrocamargo (approved):

aequilibrae/paths/AoN.pyx
```diff
@@ -285,14 +289,16 @@ def path_computation(origin, destination, graph, results):
         results.path_nodes = None
         results.path_link_directions = None
         results.milepost = None
+        results.eary_exit = False
```
jamiecook, line 292: early (not eary)

aequilibrae/paths/results/path_results.py
```diff
@@ -60,6 +60,7 @@ def __init__(self) -> None:
         self.origin = None
         self.destination = None
         self.graph: Graph = None
+        self.early_exit = False
```
jamiecook, line 63: Why do we have both `_early_exit` and `early_exit`?

aequilibrae/paths/AoN.pyx
Jake-Moss: Whoops, thanks

aequilibrae/paths/results/path_results.py
Jake-Moss: `_early_exit` is intended to privately mark the previous trees state for the path re-computation logic in `update_trace`. `early_exit` is intended to be a public configuration option, say if you want to change a `PathResults` object to use early exit on the next update trace where it previously hadn't or visa versa.

I've also realised I've change this API slightly in the A* PR to be consistent
https://github.com/AequilibraE/aequilibrae/pull/467/files#diff-2639bd28c2998e1d08e05d2939351b7cece776c19fec0eaa9c020f98e2ec33e7L295-R349

PR #467, opened 2023-11-20
Initial A* implementation

Jake-Moss:
This implementation uses a haversine distance heuristic, in theory this a consistent heuristic [1] that would give us "an optimal path without processing any node more than once" when the graph cost is set to distance, however, in testing I've found that the distance calculations based and lon/lat done by myself and QGIS do not align with those in the project. Unfortunately this makes the heuristic not admissible [2] meaning A* will not always find the least-cost path, and may not do the minimum amount of work.

From some basic instrumenting I found A* is consistently an order of magnitude better than the existing Dijkstra's implementation when it comes to heap operation counts. Time wise I haven't performed real benchmarks.

Currently there is only one heuristic implemented for distance, if the graph cost field is set to `free_flow_time` the heuristic, while not useless is theory, it is practically as the magnitudes of the distances dominates any graph cost meaning A* will explore the whole network, just like Dijkstra's.

[1] https://en.wikipedia.org/wiki/Consistent_heuristic
[2] https://en.wikipedia.org/wiki/Admissible_heuristic

### DIFF
--- a/aequilibrae/project/network/network.py
+++ b/aequilibrae/project/network/network.py
@@ -338,6 +338,8 @@ class Network(WorkerThread):
         curr.execute("select node_id from nodes where is_centroid=1 order by node_id;")
         centroids = np.array([i[0] for i in curr.fetchall()], np.uint32)
 
+        lonlat = self.nodes.lonlat.set_index("node_id")
+
         data = df[valid_fields]
         for m in modes:
             net = pd.DataFrame(data, copy=True)
@@ -346,7 +348,7 @@ class Network(WorkerThread):
             g.mode = m
             g.network = net
             if centroids.shape[0]:
-                g.prepare_graph(centroids)
+                g.prepare_graph(centroids, lonlat=lonlat)
                 g.set_blocked_centroid_flows(True)
             else:
                 get_logger().warning("Your graph has no centroids")
--- a/aequilibrae/project/network/network.py
+++ b/aequilibrae/project/network/network.py
@@ -348,10 +348,11 @@ class Network(WorkerThread):
             g.mode = m
             g.network = net
             if centroids.shape[0]:
-                g.prepare_graph(centroids, lonlat=lonlat)
+                g.prepare_graph(centroids)
                 g.set_blocked_centroid_flows(True)
             else:
                 get_logger().warning("Your graph has no centroids")
+            g.lonlat_index = lonlat.loc[g.all_nodes]
             self.graphs[m] = g
 
     def set_time_field(self, time_field: str) -> None:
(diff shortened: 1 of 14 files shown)


Jake-Moss:
Regarding the discrepancies between the lon/lat calculations and the database for Chicago Sketch, I believe we can't guarantee that the heuristic is admissible (and thus not consistent).

Here's a screenshot from from QGIS with the `distance` field compared against the derived length attribute, and one from Google maps showing roughly the same distance. Both values are within 10m of what the haversine heuristic returns,
image
image

The values from the database say that the true cost of those two links is 450m, while the distance calculations suggest it is about 640m. This means that the heuristic will overestimate the cost to reach the left most node from the right most (the estimated cost would be `0m + 330m + 182m`)
> If the heuristic function is [admissible](https://en.wikipedia.org/wiki/Admissible_heuristic) – **meaning that it never overestimates the actual cost to get to the goal** – A* is guaranteed to return a least-cost path from start to goal.

> A consistent heuristic is also [admissible](https://en.wikipedia.org/wiki/Admissible_heuristic), i.e. it never overestimates the cost of reaching the goal (the [converse](https://en.wikipedia.org/wiki/Converse_(logic)), however, is not always true).

Jake-Moss:
To come later this afternoon is a mechanism to switch heuristics at runtime to allow different cost fields like `free_flow_time`

aequilibrae/project/network/network.py
```diff
@@ -346,7 +348,7 @@ def build_graphs(self, fields: list = None, modes: list = None) -> None:
             g.mode = m
             g.network = net
             if centroids.shape[0]:
-                g.prepare_graph(centroids)
+                g.prepare_graph(centroids, lonlat=lonlat)
```
pedrocamargo, line 351: We should insert lonlat into the graph regardless of centroids here.  Otherwise it prevents the user from setting a centroid that was not coded as such in the model

Jake-Moss: Resolved in ec0117adfd5584958840d6b2581fccb8249a8d69

Jake-Moss:
I've just extended our existing path finding tests to include `early_exit` and `a_star` via subtests. Unfortunately while `unittest` (stdlib) supports them, `pytest` requires an additional package `pytest-subtests` to display them nicely.
https://github.com/pytest-dev/pytest-subtests

I haven't added A* specific predecessor tests as I think it will vary depending on the heuristic, and with such small toy networks like the triangle one I don't feel like working out the specific heuristic values and iteration order is a useful test. Particularly since the path is really all we are interested in with A*.

Jake-Moss:
Currently the `equirectangular_heuristic` is preferred over the `haversine_heuristic` as I believe the expectation with A* is that it should be faster than Dijkstra's, and from some testing on large networks with pathological OD pairs, the `haversine_heuristic` constant factors are just too large.

pedrocamargo (approved):

PR #462, opened 2023-11-03
Documentation overhaul

Jake-Moss:
Format doc strings to RST. HTML generated docs look correct. Not sure how to set the class included in the TOC.

Small change made to the noise generation, if we are creating an RNG, we should use it.

PR #473, opened 2023-12-01
Documentation for PuT (#462 fix)

Jake-Moss:


Jake-Moss:
[coquimbo_gtfs_to_graph.zip](https://github.com/AequilibraE/aequilibrae/files/13520504/coquimbo_gtfs_to_graph.zip)

@djfrancesco Here's an updated copy of the notebook I based the example off.
I've merged `develop` into `feature_PuT_assignment` so it now has the updated Coquimbo example

PR #476, opened 2023-12-05
PT Integration

Jake-Moss:
This PR is for the integration of the Python interface of the PT assignment into the existing AequilibraE structure.

My current (incomplete) plan is:
- Each PT graph is identified by a `period_id`, this is a unique key that corresponds to a period start and end within the `periods` table.
- A `Transit` object holds a dict of SF graphs (keyed by `period_id`), similar to `network.graphs`.
- A `Transit` object can create, load, and save all SF graphs
- ~~graphs are stored in the in `public_transport.sqlite` database under a unique table tables (`(links|nodes)_PERIOD_ID`) 
  This is to allow multiple graphs over different periods to exist simultaneously without having to deal with link/node id conflicts~~
- An PT assignment can be run by constructing a `TransitAssignment` object and will return a `TransitAssignmentResults` object containing aequilibrae matrix/data objects.


Jake-Moss:
I've added some base classes to facilitate code reuse and logical grouping of similar objects.

I've intended for each piece of the current traffic assignment workflow to have PT mirror
- Where `Network.graphs` stores `Graph` objects, `Transit.graphs` will store `TransitGraph`s.
- Where there's a `TrafficClass` there's an analogous `TransitClass`
- Where there's a `TrafficAssignment` there's an analogous `TransitAssignment`
- Where there's a `AssignmentResults` there's an analogous `AssignmentResultsTransit`

I've intended this to mirror the existing user interface as closely as reasonable. The existing API should not change at all and thus be backwards compatible.

These new classes are not flushed out and just for scaffolding at the moment.

aequilibrae/paths/graph_building.pyx
```diff
@@ -176,7 +176,7 @@ def build_compressed_graph(graph):
 
     df = pd.concat([df, comp_lnk])
     df = df[["id", "link_id", "a_node", "b_node", "direction"]]
-    properties = graph._Graph__build_directed_graph(df, graph.centroids)  # FIXME: Don't circumvent name mangling
+    properties = graph._build_directed_graph(df, graph.centroids)  # FIXME: Don't circumvent name mangling
```
pedrocamargo: You can remove the fix me comment now

aequilibrae/paths/traffic_assignment.py
```diff
@@ -768,3 +823,46 @@ def save_select_link_results(self, name: str) -> None:
         """
         self.save_select_link_flows(name)
         self.save_select_link_matrices(name)
+
+
+class TransitAssignment(AssignmentBase):
+    all_algorithms = ["spiess & florian", "sf"]
```
pedrocamargo: Let's rename spiess & florian to "optimal-strategies", which is what it is called in practice

aequilibrae/project/database_specification/transit/tables/periods.sql
```diff
@@ -0,0 +1,22 @@
+--@ The periods table holds the time periods and their period_id. Default entry with id 1 is the entire day.
+--#
+CREATE TABLE if not exists periods (period_id               INTEGER UNIQUE NOT NULL,
```
pedrocamargo: This table should go in the project database, as these periods are used for everything

Jake-Moss:
A the full workflow is not taking shape, the assignment results are currently wrong and inaccessible but the general shape is here now. Based on the Coquimbo example
```python
...
dest_path = join(fldr, "gtfs_coquimbo.zip")

data = Transit(project)

transit = data.new_gtfs_builder(agency="LISANCO", file_path=dest_path)

transit.load_date("2016-04-13")

transit.save_to_disk()

from aequilibrae.paths import TransitAssignment, TransitClass

graph = data.create_graph(with_outer_stop_transfers=False, with_walking_edges=False)

transit_graph = graph.to_transit_graph()

from aequilibrae.matrix import AequilibraeMatrix

zones_in_the_model = len(transit_graph.centroids)

names_list = ['pt']

mat = AequilibraeMatrix()
mat.create_empty(file_name='/tmp/path_to_matrix.aem',
                 zones=zones_in_the_model,
                 matrix_names=names_list,
                 memory_only=False)
mat.index = transit_graph.centroids[:]
mat.matrix['pt'] = np.full((zones_in_the_model, zones_in_the_model), 1.0)
mat.computational_view()

assig = TransitAssignment()

# Create the assignment class
assigclass = TransitClass(name="pt", graph=transit_graph, matrix=mat)
assig.add_class(assigclass)

assig.set_time_field("trav_time")
assig.set_frequency_field("freq")

assig.set_algorithm("os")

assig.execute()

assig.assignment._TransitGraphBuilder__classes["pt"].demand  # assig.results not implemented yet

```

Some issues I've noticed
-  To solve the issue of centroid flow `TransitGraphBuilder` (renamed from `SF_graph_builder`) makes a distinction between `origin` and `destination` nodes. Currently I'm treating both `origin` and `destination` nodes as centroids, however this goes against aequilibraes current notion because `origin` and `destination` nodes, while having different node_ids are the same node. Flow should only be allowed from `origin` to `destination`. This shouldn't be a huge issue and I plan to fix it in the coming week.
- Related to the above issue, the `compute_SF` functions assume their input is supplied in COO format in dataframe, however aequilibrae matrices are dense. Currently I have it set to convert an aequilibrae matrix to COO via scipy, then construct a DataFrame. This obviously has to go but I'm not sure the best approach. Should we convert the assignment to use a dense matrix or construct the CSR matrix the assignment constructs externally and pass it in? (@djfrancesco)
- In mimicking the existing API I've created a `TransitClass` class, while not entirely analogous to  typically traffic classes (which are dependent on the mode), these `TransitClass` classes could be constructed for a time period instead. However I'm not sure how these should be stored as any period can be constructed. I see two options for storage, store each period as an entirely separate graph with its own `links` and `nodes` table. This would be the simplest to implement but horribly inefficient. Or use a similar strategy as the existing modes use. Label each link as being present within some `TransitClass`/period then selectively load these when requested. This would require some master network similar that has all links present in it to prevent `link` and `node` id classes. Perhaps the whole day could work? Then creating a new time period/`TransitClass` would just need to label the links/nodes within the master network as being present in that time period. I'm not sure which approach would be best here. Either way I'll continue to work on the Python interface first and use the existing database saving and just require that there cannot exist more than one period/`TransitClass` at a time currently. @pedrocamargo

aequilibrae/project/database_specification/transit/tables/periods.sql
Jake-Moss: I've just move the `periods` table into the Project database. I've also created a small wrapper class around it, similar to the existing Nodes, Links, and Zones classes.

Jake-Moss:
Hey @pedrocamargo and @djfrancesco . The API is now nearing complete, and a full workflow is now possible\*. Here's the notebook example for Coquimbo. Tomorrow I plan to resolve the merge conflicts then begin writing the required tests.
[coquimbo_gtfs_to_graph.zip](https://github.com/AequilibraE/aequilibrae/files/13658790/coquimbo_gtfs_to_graph.zip)

\* The `TransitGraphBuilder` config is not saved currently its not possible to create a new `TransitGraph` object from the `.to_transit_graph()` function on a reloaded `TransitGraphBuilder` without manually setting the correct configuration options. I'll fix this tomorrow.

aequilibrae/transit/transit_graph_builder.py
```diff
@@ -810,6 +820,8 @@ def create_connector_edges(self, method="overlapping_regions", allow_missing_con
                 connectors.append(df)
             access_connector_edges = pd.concat(connectors).rename(columns={"node_id": "a_node"}).reset_index(drop=True)
 
+            breakpoint()
```
pedrocamargo: What is this?

pedrocamargo:
> A the full workflow is not taking shape, the assignment results are currently wrong and inaccessible but the general shape is here now. Based on the Coquimbo example
> 
> ```python
> ...
> dest_path = join(fldr, "gtfs_coquimbo.zip")
> 
> data = Transit(project)
> 
> transit = data.new_gtfs_builder(agency="LISANCO", file_path=dest_path)
> 
> transit.load_date("2016-04-13")
> 
> transit.save_to_disk()
> 
> from aequilibrae.paths import TransitAssignment, TransitClass
> 
> graph = data.create_graph(with_outer_stop_transfers=False, with_walking_edges=False)
> 
> transit_graph = graph.to_transit_graph()
> 
> from aequilibrae.matrix import AequilibraeMatrix
> 
> zones_in_the_model = len(transit_graph.centroids)
> 
> names_list = ['pt']
> 
> mat = AequilibraeMatrix()
> mat.create_empty(file_name='/tmp/path_to_matrix.aem',
>                  zones=zones_in_the_model,
>                  matrix_names=names_list,
>                  memory_only=False)
> mat.index = transit_graph.centroids[:]
> mat.matrix['pt'] = np.full((zones_in_the_model, zones_in_the_model), 1.0)
> mat.computational_view()
> 
> assig = TransitAssignment()
> 
> # Create the assignment class
> assigclass = TransitClass(name="pt", graph=transit_graph, matrix=mat)
> assig.add_class(assigclass)
> 
> assig.set_time_field("trav_time")
> assig.set_frequency_field("freq")
> 
> assig.set_algorithm("os")
> 
> assig.execute()
> 
> assig.assignment._TransitGraphBuilder__classes["pt"].demand  # assig.results not implemented yet
> ```
> 
> Some issues I've noticed
> 
> * To solve the issue of centroid flow `TransitGraphBuilder` (renamed from `SF_graph_builder`) makes a distinction between `origin` and `destination` nodes. Currently I'm treating both `origin` and `destination` nodes as centroids, however this goes against aequilibraes current notion because `origin` and `destination` nodes, while having different node_ids are the same node. Flow should only be allowed from `origin` to `destination`. This shouldn't be a huge issue and I plan to fix it in the coming week.
> * Related to the above issue, the `compute_SF` functions assume their input is supplied in COO format in dataframe, however aequilibrae matrices are dense. Currently I have it set to convert an aequilibrae matrix to COO via scipy, then construct a DataFrame. This obviously has to go but I'm not sure the best approach. Should we convert the assignment to use a dense matrix or construct the CSR matrix the assignment constructs externally and pass it in? (@djfrancesco)
> * In mimicking the existing API I've created a `TransitClass` class, while not entirely analogous to  typically traffic classes (which are dependent on the mode), these `TransitClass` classes could be constructed for a time period instead. However I'm not sure how these should be stored as any period can be constructed. I see two options for storage, store each period as an entirely separate graph with its own `links` and `nodes` table. This would be the simplest to implement but horribly inefficient. Or use a similar strategy as the existing modes use. Label each link as being present within some `TransitClass`/period then selectively load these when requested. This would require some master network similar that has all links present in it to prevent `link` and `node` id classes. Perhaps the whole day could work? Then creating a new time period/`TransitClass` would just need to label the links/nodes within the master network as being present in that time period. I'm not sure which approach would be best here. Either way I'll continue to work on the Python interface first and use the existing database saving and just require that there cannot exist more than one period/`TransitClass` at a time currently. @pedrocamargo

Using a COO matrix for assignment is absolutely fine, as the conversion through scipy.sparse is incredibly fast.

pedrocamargo:
> Hey @pedrocamargo and @djfrancesco . The API is now nearing complete, and a full workflow is now possible*. Here's the notebook example for Coquimbo. Tomorrow I plan to resolve the merge conflicts then begin writing the required tests. [coquimbo_gtfs_to_graph.zip](https://github.com/AequilibraE/aequilibrae/files/13658790/coquimbo_gtfs_to_graph.zip)
> 
> * The `TransitGraphBuilder` config is not saved currently its not possible to create a new `TransitGraph` object from the `.to_transit_graph()` function on a reloaded `TransitGraphBuilder` without manually setting the correct configuration options. I'll fix this tomorrow.

I like what I see here.  Let's update the examples that ship with the model to have all the tables we need (not criitical)

aequilibrae/transit/transit_graph_builder.py
djfrancesco: I only use `print('got this far')` to debug

Jake-Moss: Ah whoops  I thought I'd caught all those

Jake-Moss: Hmm looks like this is a from an older commit

djfrancesco (approved):

aequilibrae/paths/public_transport.pyx
djfrancesco, line 1: There are many modifications here, making it a bit challenging for me to provide a quick review. I trust you in making these changes to align with the project.

djfrancesco (approved):

docs/source/examples/full_workflows/plot_public_transit_assignment.py
```diff
@@ -117,38 +127,60 @@
 # %%
 # Mock demand matrix
 # ~~~~~~~~~~~~~~~~~~
-# We'll create a mock demand martix with demand `1` for every zone.
+# We'll create a mock demand matrix with demand `1` for every zone.
 # We'll also need to convert from `zone_id`s to `node_id`s.
```
pedrocamargo, line 131: @Jake-Moss  & @djfrancesco, what is this?  Aren't we using zone IDs?  If not, that makes it VERY difficult to use the software

Jake-Moss: Maybe I've worded this comment poorly. AFAIU in aequilibre demand matrices are done centroid to centroid. The transit graph builder takes zones and constructs centroids in them. These are either "od" or "origin" and "destination" node types. The Transit assignment deals with blocking centroid flow by making this distinction (without blocking flow only "od"s are constructed). In the `to_transit_graph` method these centroids are set as the centroids, but because AequilibraE matrices only have a single index between both rows and cols we translate the node ids of the centroids (which are only "od"s regardless of centroid flow) to the corresponding node ids of the "destination"/"origin" node for the assignment

Jake-Moss: The relevant pieces of code are
- `create_od_vertices` https://github.com/Jake-Moss/aequilibrae/blob/8798f47e4f69ae0f9da46a3a4cbbfc096f3d6f92/aequilibrae/transit/transit_graph_builder.py#L503
- `to_transit_graph` https://github.com/Jake-Moss/aequilibrae/blob/8798f47e4f69ae0f9da46a3a4cbbfc096f3d6f92/aequilibrae/transit/transit_graph_builder.py#L1415
- `create_od_node_mapping` https://github.com/Jake-Moss/aequilibrae/blob/8798f47e4f69ae0f9da46a3a4cbbfc096f3d6f92/aequilibrae/transit/transit_graph_builder.py#L544
- Optimal strategies https://github.com/Jake-Moss/aequilibrae/blob/8798f47e4f69ae0f9da46a3a4cbbfc096f3d6f92/aequilibrae/paths/optimal_strategies.py#L33-L38

djfrancesco: Yes, as @Jake-Moss explained, the demand matrix is expressed in terms of zone IDs, while the assignment method is only dealing with demand between graph vertex IDs. This is why there is the `create_od_node_mapping` method, to get the vertex-based matrix from the zone-based one. This method belongs to the `transit_graph_builder` because this is where we have all the information, of zone IDs and their respective centroids. 
Now in order to block centroid flows, we duplicate the zone centroids into distinct origin and destination nodes. Access connectors start from origin nodes, while egress connectors ends at destination nodes. These origin and destination nodes from a same zone are not directly connected in the graph, which is why this is blocking centroid flows.

pedrocamargo (approved):

PR #478, opened 2023-12-07
removes use of necessary sqlite3 connection cursors and cached connections

pedrocamargo:


jamiecook (approved):
This is one where i'm pretty glad we've got good tests. 
Looking through it all seems to be fairly straightforward - hard to tell for sure without running everything though.

PR #492, opened 2024-01-11
Route choice

Jake-Moss:
This pr aims to implemented the BFS-LE algorithm as described in Rieser-Schüssler, Balmer, and Axhausen,
'Route Choice Sets for Very High-Resolution Data'.

Further notes and such are included at the top of the `bfs_le.pyx` file.
The method is implemented as a Cython extension with an extension class.

This is an incredibly rough implementation. It makes heavy use of libcpp data structures and blatantly makes inefficient copies and allocations. 

Todo:
- [X] Basic functionality, results size and depth limiting
- [x] Instead of sampling routes to fill the rest of the route set, sample the removed link sets early to prevent unnecessary work. (PO1)
- [X] Migrate A* pathfinding to compressed graph. (P02 [sic])
- [x] Benchmarking
- [x] ~~Implemented custom trie data structures for more efficient route and removed link set storage (requires more thought currently)~~
- [x] Tests
- [x] ~~Parallelise~~ I believe this method should be parallelised at a higher level, it has just enough moving parts and a clear parallel implementation isn't obvious to me.
- [x] Rename files to be route choice themed, `bfs_le.pyx` isn't a great name
- [x] Documentation

In testing I found an issue with the A* implementation, because it most revisit nodes if it believes it over estimated them but hasn't found the destination yet, the `reached_first` array may be overflown. I've removed this until if/when I can revisit this to implement proper skimming for A*.


aequilibrae/paths/route_choice.pyx
```diff
@@ -0,0 +1,253 @@
...
+            unordered_map[unordered_set[long long] *, vector[long long] *] *working_set
+            unordered_map[unordered_set[long long] *, vector[long long] *].const_iterator set_iter
+            vector[unordered_set[long long] *] queue
+            vector[unordered_set[long long] *] next_queue
+            unordered_set[long long] *banned
+            unordered_set[long long] *new_banned
+            unordered_set[vector[long long] *] *route_set
+            vector[long long] *vec
+            pair[unordered_set[long long] *, vector[long long] *] *pair_tmp
+            pair[unordered_set[long long] *, vector[long long] *] x
```
Jake-Moss: All of these maps/sets of pointers make no sense. A pointer will be unique for each object we construct. Looks like I might have to implement my own hash functions after all, or custom data structure

Jake-Moss:
image

Based on a rudimentary profiling I'm not sure further optimisation here via custom data structures is wise. Path finding, as expected, is taking the vast majority of the runtime.

Tomorrow I'll have some graphs and proper benchmarks.

Jake-Moss:
From todays bug fixes and performance improvements the runtime when generating 1000 unique routes over the densest part of the Arkansas model takes on average 2.1s single threaded on my Ryzen 5600x. On average these routes are 352.108 links long. BFSLE reached a depth of 4.
MicrosoftTeams-image

More profiling indicates no real change in where the time is spent. Path finding is still by far the slowest portion of this method.
image

Jake-Moss:
I'm marking this as ready to review as I think it's functionally complete. Tests to come tomorrow.

Jake-Moss:
Looking at the annotated Cython output doesn't look great. Unfortunately not much can be done without writing custom cdef extern's for the stdlib and lying to the Cython compiler that these functions don't throw exceptions to prevent it from attempting to acquire the gil.

However, the gcc docs assure me that the overhead of the cpp exceptions themselves, particularly the try-catch blocks, is about 7% and and that gcc handles these in an efficient manner. So if an exception is never thrown, the gil is never acquired, and the overhead of the try-catch blocks is negligible. If something ever does throw an exception anyway we have bigger problems.

image

Jake-Moss:
After review I plan to merge this into the new feature branch `route_choice` and use that as the target for new PRs.

pedrocamargo:
@janzill , Can you merge this into the feature branch in case you are happy with the BFS-LE?

pedrocamargo (approved):

pedrocamargo:
> From todays bug fixes and performance improvements the runtime when generating 1000 unique routes over the densest part of the Arkansas model takes on average 2.1s single threaded on my Ryzen 5600x. On average these routes are 352.108 links long. BFSLE reached a depth of 4. MicrosoftTeams-image
> 
> More profiling indicates no real change in where the time is spent. Path finding is still by far the slowest portion of this method. image

@Jake-Moss , can you create this same map but using Dijkstra instead? Just for documentation purposes

Jake-Moss:
> > From todays bug fixes and performance improvements the runtime when generating 1000 unique routes over the densest part of the Arkansas model takes on average 2.1s single threaded on my Ryzen 5600x. On average these routes are 352.108 links long. BFSLE reached a depth of 4. MicrosoftTeams-image
> > More profiling indicates no real change in where the time is spent. Path finding is still by far the slowest portion of this method. image
> 
> @Jake-Moss , can you create this same map but using Dijkstra instead? Just for documentation purposes

Both Dijkstra's and A* are producing the exact same results, this is likely because the heuristic uses is consistent. However I have added an option to switch between A* and Dijkstra's. Changing the heuristic used isn't current possible with some hacks due to redefinitions errors caused by the path finding module being `included`

Jake-Moss:
I've opened a dummy PR containing some scratch work for saving route choice results while this PR is in review, once it's merged I'll move that PR over to the main repo. https://github.com/Jake-Moss/aequilibrae/pull/1

janzill (approved):
as discussed, looking good

PR #495, opened 2024-01-27
Changes OSM downloader to accept a polygon instead of a bounding box

pedrocamargo:
This PR allows users to provide a polygon instead of a bounding box when downloading from OSM. This is a substantial improvement for cases where the geometry of the model is irregular.


Jake-Moss (approved):
lgtm, just removed some indentation related to the previous threshold check.

PR #496, opened 2024-01-29
new matrix API

pedrocamargo:


aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -122,6 +126,38 @@ def save(self, names=()) -> None:
...
+            mat.create_empty(**args)
+            mat.indices[:, :] = self.indices[:, :]
+            for core in cores:
+                mat.matrix[core][:, :] = self.matrix[core][:, :]
+            mat.name = self.name
+            mat.description = self.description
+            mat.close()
+            del mat
+
+        elif ".omx" in file_name:
```
Jake-Moss: Ideally we'd be using pathlib's `suffix` to handle this. A path can contain `.omx` or `.aem` somewhere that's not the extension. One day we should move over to just using pathlib everywhere to avoid messy strings 
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix

Jake-Moss (approved):

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: Agreed and fixed

Issue #493, opened 2024-01-16
Discrepancy between select link calculated flows and assigned flows for CFW/BFW


mitsudz:
**Purpose of use**
Problem found during SMP internship working on the ODME algorithm.

**Describe the bug**
When running an assignment with either conjugate franke-wolfe or biconjugate frank-wolfe, the select link analysis and assignment flow to not match up. In particular - following select link analysis we obtain a matrix which contains in each cell the amount of demand through the link originating from the corresponding origin-destination pair. Summing these up in theory should give the flow assigned by the algorithm for a given link-direction - however this differs from the results obtained from 
the assignment. This does not occur for algorithms other than cfw & bfw, and the discrepancy only occurs when we set the maximum number of iterations to > 1.

**To Reproduce**
Got to the odme branch at https://github.com/mitsudz/aequilibrae/tree/mitsudz/odme. Navigate to tests/aequilibrae/paths and see the script titled S_L_Problem_Script.ipynb. From here you can see code to create an example assignment object and run it with cfw (the same behaviour happens with bfw) with max iterations of 5 and check 2 selected links. The results there are shown in the screenshots below, and you can see the assigned and calculated flows differ in both cases.

**Screenshots**
<img width="517" alt="Screenshot 2024-01-16 at 4 21 11 pm" src="https://github.com/AequilibraE/aequilibrae/assets/86662260/cb3e3045-363d-4fb5-815c-696c18c24d43">
<img width="666" alt="Screenshot 2024-01-16 at 4 21 21 pm" src="https://github.com/AequilibraE/aequilibrae/assets/86662260/0b419eec-888d-4ad6-9886-bea8bd00cd25">

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.2.1
 - Python version 3.10.12
 - Version - AequilibraE 0.9.4 (develop) - release name: Queluz

pedrocamargo:
Wasn't this fixed, @Jake-Moss ?

Jake-Moss:
Whoops, yes this was

PR #500, opened 2024-01-31
Select link cwf bfw bug

Jake-Moss:
Addressing the bug report raised in https://github.com/AequilibraE/aequilibrae/issues/493

Current thoughts are that the blending of the previous and current iteration is failing. The fix in 4894f619ff2393ad0244d44db8be8343d57748e5 doesn't seem to be enough, though it does extend the number of iterations where the results agree from 1 to 4.

### DIFF
--- a/aequilibrae/paths/linear_approximation.py
+++ b/aequilibrae/paths/linear_approximation.py
@@ -133,23 +133,12 @@ class LinearApproximation(WorkerThread):
 
         if self.algorithm in ["cfw", "bfw"]:
             for c in self.traffic_classes:
-                r = AssignmentResults()
-                r.prepare(c.graph, c.matrix)
-                r.compact_link_loads = np.zeros([])
-                r.compact_total_link_loads = np.zeros([])
-                self.previous_step_direction[c._id] = r
-
-                r = AssignmentResults()
-                r.prepare(c.graph, c.matrix)
-                r.compact_link_loads = np.zeros([])
-                r.compact_total_link_loads = np.zeros([])
-                self.step_direction[c._id] = r
-
-                r = AssignmentResults()
-                r.prepare(c.graph, c.matrix)
-                r.compact_link_loads = np.zeros([])
-                r.compact_total_link_loads = np.zeros([])
-                self.pre_previous_step_direction[c._id] = r
+                for d in [self.step_direction, self.previous_step_direction, self.pre_previous_step_direction]:
+                    r = AssignmentResults()
+                    r.prepare(c.graph, c.matrix)
+                    r.compact_link_loads = np.zeros([])
+                    r.compact_total_link_loads = np.zeros([])
+                    d[c._id] = r
 
     def calculate_conjugate_stepsize(self):
         self.vdf.apply_derivative(
@@ -247,7 +236,6 @@ class LinearApproximation(WorkerThread):
         # 2nd iteration is a fw step. if the previous step replaced the aggregated
         # solution so far, we need to start anew.
         if self.iter == 2 or self.stepsize == 1.0 or self.do_fw_step or self.algorithm in ["msa", "frank-wolfe"]:
-            print("hit iter 2 step dir block")
             self.do_fw_step = False
             self.do_conjugate_step = True
             self.conjugate_stepsize = 0.0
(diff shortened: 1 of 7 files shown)


aequilibrae/paths/linear_approximation.py
```diff
@@ -257,13 +260,29 @@ def __calculate_step_direction(self):
...
         # 3rd iteration is cfw. also, if we had to reset direction search we need a cfw step before bfw
         elif (self.iter == 3) or (self.do_conjugate_step) or (self.algorithm == "cfw"):
+            print("hit iter 3 step dir block")
             self.do_conjugate_step = False
             self.calculate_conjugate_stepsize()
             for c in self.traffic_classes:
                 sdr = self.step_direction[c._id]
                 pre_previous = self.pre_previous_step_direction[c._id]
+
                 copy_two_dimensions(pre_previous.link_loads, sdr.link_loads, self.cores)
```
Jake-Moss: Additional another bug may have been introduced in this commit 21c66f117fe2c28bedfec00afba54e27c517635a (thanks @janzill ). The current step direction is being copied into the `pre_previous` matrix, this is used to hold the step direction before it is copied back out into the `previous` matrix. However this copy never occurs. I believe it should be copied directly into the `previous` matrix instead. 

If the next iteration is `fw` or `msa` due to restarting, then  `pre_previous`  is not accessed.
If the next iteration is `cfw` then  `pre_previous`  is immediately over written again.
If the next iteration is `bfw` then the  `pre_previous`  is immediately over written.

` pre_previous` is never read from without before being written to.
If a `cfw` then a `bfw` iteration occurs then the `previous` matrix may not have been written to before hand.

janzill: Yes, I agree. I think this actually is a bug when running BFW because we use previous_step_direction in finding the blend coefficients (betas) in the BFW, which is run before any copying happens. Also, pre_previous is a bit of a misnomer because it is only ever used in BFW to hold a temporary copy during step direction calculations

Jake-Moss:
Overall the root of the issue was that the select link wasn't taking into account the previous step directions in it's
calculations. In principle, the patch that modified the select to do just that should have worked, however, the matrix
that stored the previous step directions was allocated on the wrong object and, while the object lived the entire
duration of the assignment, the matrix was being zeroed at each iteration. I'd overlooked this as it did improve the
stability up to 4 iterations, but Jan pointed out to me that the beta's being used were 1.0 (and 0.0 respectively) up
until the 5th iteration where the step direction truly started to matter.

Another two bugs I found are patched.

1. The aon object was being leaked outside of the traffic class for loop, this meant that the last traffic class was
being used for all select links analysis'. The aon objects are now stored and reused, this may increase memory usage
during the assignment as GC can no longer clean the memory allocated by these objects up at the end of every AON
assignment, instead it will at the end of all assignments.

2. As noted in the comment `pre_previous_step_direction` was being used in place of `previous_step_direction`.

aequilibrae/paths/linear_approximation.py
Jake-Moss: This should be resolved, though I did not change the name

pedrocamargo (approved):

aequilibrae/paths/linear_approximation.py
```diff
@@ -124,30 +124,21 @@ def __init__(self, assig_spec, algorithm, project=None) -> None:
         self.previous_step_direction = {}  # type: Dict[AssignmentResults]
         self.pre_previous_step_direction = {}  # type: Dict[AssignmentResults]
```
janzill: Can we rename `pre_previous_step_direction` to `temp_step_direction_for_copy` or something like that? It's only ever used as a temporary place to store step_direction results because we operate on it in-place and need the original data in the iteration

janzill: can make this a separate PR too, but it fits with the small bug fix

janzill (approved):
lgtm, nice catch on the cfw pre_previous bug. To clarify, this should not influence results much because it only matters when running bfw and then only on the first bfw step-direction calculation after a cfw step (the latter has to happen when building up the previous and pre-previous step direction, so only once intitially or whenever the step-direction resets).

aequilibrae/paths/linear_approximation.py
Jake-Moss: Just renamed those as well as the others I added.

PR #506, opened 2024-02-05
OSM importer speedup

pedrocamargo:
These speed-ups are going to be incredibly useful in building models automatically for Polaris and when creating models for large areas.


aequilibrae/project/network/osm/osm_builder.py
```diff
@@ -1,263 +1,239 @@
...
-            raise e
-
-    def __worksetup(self):
-        self.__link_types = self.project.network.link_types
-        lts = self.__link_types.all_types()
-        for lt_id, lt in lts.items():
-            self.__model_link_types.append(lt.link_type)
-            self.__model_link_type_ids.append(lt_id)
+        del self.node_df
+        gc.collect()
```
Jake-Moss, line 148: Are these calls to `gc.collect()` shown to reduce memory usage? If we've derived something from this df then ensuring Python gets that memory back might be a little trickery that just `del` and `gc.collect()`

I don't believe this would be hurting us in any manner, more like that it might just be unneeded.

aequilibrae/project/network/osm/osm_builder.py
```diff
@@ -1,263 +1,239 @@
...
+        del self.links_df
+        gc.collect()
+        self.logger.info("Adding links to file")
+        self.__emit_all(["text", "Adding links to file"])
+        conn.executemany(insert_qry, links_df)
+
+    def _build_geometry(self, nodes: List[int]) -> str:
+        slice = self.node_df.loc[nodes, :]
+        txt = ",".join((slice.lon.astype(str) + " " + slice.lat.astype(str)).tolist())
+        return f"LINESTRING({txt})"
```
Jake-Moss, line 175: Is this to avoid shapely object construction?

aequilibrae/project/network/osm/osm_builder.py
```diff
@@ -1,263 +1,239 @@
...
-        if inside_area == 0:
-            return None
-        geometry = "LINESTRING ({})".format(", ".join(geometry))
-
-        attributes = [vars.get(x) for x in fields]
-        attributes.append(geometry)
-        return attributes
-
-    def __repair_link_type(self, link_type: str) -> str:
+    def __define_link_type(self, link_type: str) -> [str, str]:
```
Jake-Moss: I'm not sure this is a valid type hint, I would have expected it to be
```python
   def __define_link_type(self, link_type: str) -> tuple[str, str]:
        ...
        return letter, link_type
```
or
```python
   def __define_link_type(self, link_type: str) -> list[str]:
```

pyproject.toml
```diff
@@ -6,7 +6,7 @@ extend-exclude = '''docs/*'''
 
 
 [build-system]
-requires = ["setuptools", "numpy", "cython", "pyaml", "pyqt5", "requests", "scipy", "shapely", "pandas", "pyarrow", "pyproj", "wheel"]
+requires = ["setuptools", "numpy", "cython", "pyaml", "pyqt5", "requests", "scipy", "shapely", "pandas", "pyarrow", "pyproj", "wheel", "geopandas"]
```
Jake-Moss: Does the build system require geopandas? I don't think any cython methods require it? I'm also not 100% on some others here as well

```suggestion
requires = ["setuptools", "numpy", "cython", "pyaml", "pyqt5", "requests", "scipy", "shapely", "pandas", "pyarrow", "pyproj", "wheel"]
```

Jake-Moss (commented):
Generally looks good to me though I'm not familiar is the OSM networks. The gist of the changes it does seem like it would be faster, a lot of python loops have been moved into pandas magic or similar

pyproject.toml
pedrocamargo: Most are not required indeed. I think I fixed that in the CI work.

aequilibrae/project/network/osm/osm_builder.py
pedrocamargo: Yes.  converting to shapely, then wkb and then spatialite is unecessary.

Jake-Moss (approved):
Looks good to me

PR #507, opened 2024-02-06
Improves dependencies

pedrocamargo:
OpenMatrix is a full dependency now, so some old code could go.


pedrocamargo:
This one is trivial...

janzill (approved):

PR #515, opened 2024-03-01
Route choice API, link loading, path file generation

Jake-Moss:
TODO:
- [X] Basic path file computation
- [X] Basic link flow computation
- [X] Adjust existing tests to new API
- [x] API solidification, it should mirror existing APIs. I think I'll go with a thin wrapper around the existing route choice object
- [x] Support multiple matrix slices. Current implementation of link loading assumes a single slice, this isn't difficult to change
- [x] Separate path file generation and link loading for now. Currently the link loading implementation uses the path files as an intermediate result
- [x] Map all outputs back to network IDs, currently everything is in compact form
- [ ] ~Saving of sparse path files into parquet~
- [x] Saving of dense and sparse link flows into results database
- [x] New tests for user facing class
- [ ] General code quality improvements, some methods are deeply intertwine with each other and generally too long, particularly the `batched` method, which is 252 lines 
- [x] Documentation example


pedrocamargo:
@janzill , can you take a look at this PR sometime before next Tuesday?  That would allow Jake to jump straight into it

aequilibrae/paths/graph.py
```diff
@@ -505,6 +512,71 @@ def save_compressed_correspondence(self, path, mode_name, mode_id):
         node_path = join(path, f"nodes_to_indices_c{mode_name}_{mode_id}.feather")
         pd.DataFrame(self.nodes_to_indices, columns=["node_index"]).to_feather(node_path)
 
+    def create_compressed_link_network_mapping(self):
```
pedrocamargo, line 537: Doesn't this already exist in the the graph and network tables?  A   **__compressed_id__** field or something of that sort?

Jake-Moss: This is a one-to-many map of compressed IDs to `link_id`s that can be accessed without using a groupby or other python related query.
For each compressed link ID, `idx[id]:idx[id + 1]` is the range of indices of `data` that contains the `link_id`s that the compressed link `id` corresponds to. It also sorts them in the correct manner

There is an existing mapping but I'm not sure exactly what it accomplishes, I think it maps graph link to network links. I don't believe it is for the compressed graph 
https://github.com/AequilibraE/aequilibrae/blob/6715319a6845be1de7ab015e074b963db84ba870/aequilibrae/paths/results/assignment_results.py#L251-L260

aequilibrae/paths/route_choice_set.pyx
```diff
@@ -274,26 +327,35 @@ cdef class RouteChoiceSet:
         cdef:
             RouteSet_t *route_set
             pair[vector[long long] *, vector[long long] *] freq_pair
-            vector[long long] *link_union = <vector[long long] *>nullptr
+            vector[long long] *link_union_scratch = <vector[long long] *>nullptr
+            vector[vector[long long] *] *link_union_set = <vector[vector[long long] *] *>nullptr
             vector[vector[double] *] *cost_set = <vector[vector[double] *] *>nullptr
             vector[vector[double] *] *gamma_set = <vector[vector[double] *] *>nullptr
```
janzill: Can we rename this so it's clear what gamma stands for? path_overlap or something like that?

Jake-Moss: Done

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+from typing import Optional, Union, Tuple, List
+import pyarrow as pa
+import pathlib
+import itertools
+
+import logging
+
+
+class RouteChoice:
+    all_algorithms = ["bfsle", "lp", "link-penalisation"]
```
pedrocamargo: Let's add the American spelling too

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+    def set_paramaters(self, par: dict):
+        """
+        Sets the parameters for the route choice  TODO, do we want link specific values?
+
+        "beta", "theta", and "seed" are BFSLE specific parameters and will have no effect on link penalisation.
+        "penalty" is a link penalisation specific parameter and will have no effect on BFSLE.
+
+        Setting `max_depth`, while not required, is strongly recommended to prevent runaway algorithms.
+
+        - When using BFSLE `max_depth` corresponds to the maximum height of the graph of graphs. It's value is
```
pedrocamargo: Let's add the link to the BFSLE paper here

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+            is a recommended starting point. For large networks a good starting value is 5. Increase the value
+            until the number of desired routes is being consistently returned.
+
+        - When using LP, `max_depth` corresponds to the maximum number of iterations performed. While not enforced,
+            it should be higher than `max_routes`. It's value is dependent on the magnitude of the cost field,
+            specifically it's related to the log base `penalty` of the ratio of costs between two alternative routes.
+
+
+        Parameter values can be scalars (same values for the entire network) or network field names
+        (link-specific values) - Examples: {'alpha': 0.15, 'beta': 4.0} or  {'alpha': 'alpha', 'beta': 'beta'}
```
pedrocamargo: How (or which) parameters would be link-specific?

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+                    raise ValueError("Route choice results not computed and read/save path not specificed") from err
+                self.results = pa.dataset.dataset(
+                    self.where, format="parquet", partitioning=pa.dataset.HivePartitioning(self.schema)
+                )
+
+        return self.results
+
+    def get_load_results(
+        self, which: str = "uncompressed"
+    ) -> Union[Tuple[AequilibraeData, AequilibraeData], Tuple[AequilibraeData]]:
```
pedrocamargo: Let's replace AequilibraEData with PandasDataFrame.  I want to start deprecating AequilibraEData in favor of DFs and Parquet/Feather

pedrocamargo:
@Jake-Moss , the API looks pretty good.  I think that with some minor tweaks and the example notebook, it will be perfect.

aequilibrae/paths/route_choice.py
Jake-Moss: Whoops, left one comments from a copy paste. We could have link specific parameters, say for link penalisation, but I'm not sure how useful that would be.

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+        self.psl_schema = RouteChoiceSet.psl_schema
+
+        self.compact_link_loads: Optional[np.array] = None
+        self.link_loads: Optional[np.array] = None
+        self.results: Optional[pa.Table] = None
+        self.where: Optional[pathlib.Path] = None
+
+        self.nodes = Optional[Union[List[int], List[Tuple[int, int]]]] = None
+
+    def set_algorithm(self, algorithm: str):
```
pedrocamargo: When we talk about route choice, the algorithm in question is usually the one taking care of the route overlap itself (e.g. Path-Size Logit, Remulla, C-Logit, etc.), so let's change this to **set_choice_set_generation** and merge *set_parameters* in there with it, as that's all about the same thing.

aequilibrae/paths/route_choice.py
```diff
@@ -0,0 +1,293 @@
...
+        elif isinstance(nodes[0], int):
+            self.nodes = list(itertools.permutations(nodes, r=2))
+
+    def execute_single(self, origin: int, destination: int):
+        if self.__rc is None:
+            self.__rc = RouteChoiceSet(self.graph)
+
+        return self.__rc.run(origin, destination, **self.paramaters)
+
+    def execute(self, path_size_logit: bool = False):
```
pedrocamargo: Let's change the name of the parameter to **perform_assignment** *=True*.

Eventually, we will have other route choice algorithms in there (e.g. remulaa), and then we add a method for *set_algorithm*

Jake-Moss:
Didn't realise `dict | dict` was a 3.9 feature, I'll fix that next week

pedrocamargo:
> Didn't realise `dict | dict` was a 3.9 feature, I'll fix that next week

We no longer support Python 3.8, though

pedrocamargo:
@Jake-Moss , in order to save choice sets to disk, we would need to either save the relationship between compact graph link and link/direction or convert it to link_id/direction.

Compact graphs are not permanent, so this becomes dangerous when saving to disk


image

Jake-Moss:
> @Jake-Moss , in order to save choice sets to disk, we would need to either save the relationship between compact graph link and link/direction or convert it to link_id/direction.
> 
> Compact graphs are not permanent, so this becomes dangerous when saving to disk
> 
> image

What file is that from? I can't see it on the latest commit, I thought that doc string was modified, either way any doc string that mentions the compact links is out of date, except `apply_link_loading`, which is used by `get_load_results` to  return both
https://github.com/AequilibraE/aequilibrae/blob/f55eac93e6c051330f66e56bfdf1af0334d019bf/aequilibrae/paths/route_choice.py#L330-L361
and the translation for the results table happens during its construction
https://github.com/AequilibraE/aequilibrae/blob/f55eac93e6c051330f66e56bfdf1af0334d019bf/aequilibrae/paths/route_choice_set.pyx#L1173-L1182

Jake-Moss:
I believe this is largely ready now, it is lacking select link analysis support, though that will come in a separate PR

tests/aequilibrae/paths/test_route_choice.py
```diff
@@ -161,43 +171,126 @@ def test_round_trip(self):
...
-        table = table.to_pandas()
+        rc.batched(nodes, max_routes=20, max_depth=10, path_size_logit=True)
+        table = rc.get_results().to_pandas()
 
         gb = table.groupby(by=["origin id", "destination id"])
         for od, df in gb:
-            self.assertAlmostEqual(1.0, sum(df["probability"].values), msg="Probability not close to 1.0")
+            self.assertAlmostEqual(1.0, sum(df["probability"].values), msg=", probability not close to 1.0")
+
+    def test_link_loading(self):
```
janzill, line 219: Can we add a test where we know the answer, like for one od-pair and demand 1, where the link flows should should match the sum of probabilities across paths in each route set? We can use a small number of paths, max_depth=3 maybe.

janzill (approved):
looks great, one small suggestion regarding testing assignment

tests/aequilibrae/paths/test_route_choice.py
Jake-Moss: I've added two tests with known link loading results, one which has no demand between the selected OD pairs, and one with demand of 1 between single pair over a single link. The difference in cost is large enough that the probabilites come out binary so I've just asserted against 0.0 and 1.0, perhaps we should normalise the link costs so the probabilies are less extreme

Jake-Moss:
Hmm I thought the CI was fixed...

pedrocamargo:
Latest Mac-os is now M1...

Jake-Moss:
> Latest Mac-os is now M1...

Hmm that'll do it

Jake-Moss:
> Latest Mac-os is now M1...

I'm not sure what's going on now, the logs files said that the system was build a `x86_64` wheel running on `arm` but `deallocate` doesn't seem have had an update in the last couple months. I wouldn't have expected that to break it. We'll see how this CI runs, testing locally gave me nothing to go on.

It doesn't seem to be respecting the `--ignore-missing-dependencies``

Jake-Moss:
This PR is long enough, going to merge into the feature branch and merge develop, hopefully that'll fix some of the CI

PR #524, opened 2024-05-08
Use flat layout default excludes to skip the test dir & Fix duplicate values in gmns mode creation

Jake-Moss:


pedrocamargo (approved):

Jake-Moss:
I've largely fixed these issues, had to bump the doc building python version. Also added doc tests via pytest but disabled for now. Need to separate the latex and html doc builds due to tmp file conflicts. A couple docs tests also fail due to exceptions in the demo code, I've caught a handful (not pushed), will get the rest tomorrow morning

Jake-Moss:
I attempted to fix the documentation CI today and in the morning enabled some doc tests. However, the doc strings were not written with doc tests in mind and would require a very large overhaul to function. I've enabled it for just 3 files for a jumping off place in the future.

I also attempted to fix the documentation and I'm not sure of the `gmns_builder.py` code base. The sqlite error is genuine, it's not limited to the LaTeX build or a carry over issue. However I was rather hesitant to change anything, it's doing a lot more than I understand

pedrocamargo (approved):

PR #525, opened 2024-05-10
PT preloading implementation

### LINKED ISSUE #343: Add preloading field for traffic assignment
I don't know if it's a library limitation or just a GUI function to add : 
- Can allow for example bicycle assignment with specific VD parameters after "traffic" assignment
- Will also help to use PT vehicles as preloading for shared links

mitsudz:
Development of public transport preload integration into TrafficAssignment

closes #343


aequilibrae/project/network/network.py
```diff
@@ -426,3 +427,44 @@ def __count_items(self, field: str, table: str, condition: str) -> int:
         with commit_and_close(connect_spatialite(self.project.path_to_file)) as conn:
             c = conn.execute(f"select count({field}) from {table} where {condition};").fetchone()[0]
         return c
+
+    def build_pt_preload(
+            graph, start_time: int, end_time: int, default_pce: float = 1.0, inclusion_cond: str ="start"
+            ) -> dict[(int, int), int]:
```
jamiecook: ideally this should be a `np.array` of the correct size to be immediately summable with the load vector created during assignment.

mitsudz: Changed to this

aequilibrae/paths/linear_approximation.py
```diff
@@ -114,6 +114,9 @@ def __init__(self, assig_spec, algorithm, project=None) -> None:
 
         # Instantiates the arrays that we will use over and over
         self.capacity = assig_spec.capacity
+        if assig_spec.apply_pt_preload:
```
pedrocamargo: Nope...  Assignment is a load, not a deduction in capacity.

There is the obvious issue that pre-load could be larger than a link capacity (unlikely, but possible)

And this shifts the volume delay function, which is highly non-linear

pedrocamargo:
There is a fundamental problem with how the pre-loading is done, which I mentioned in a comment.  
Aside from that, I would treat preloads as a single class, with all preloads already there (PT is not the only possible one, as there are multiple urban services that could be considered - e.g. garbage collection, sweeping, etc.). We would not assign them, obviously, but we would include them in the line search and all other portions of the assignment, including results.

docs/source/examples/creating_models/plot_import_gtfs.py
```diff
@@ -9,20 +26,27 @@
...
+from pathlib import Path # Check filepaths...
+
 import folium
 import pandas as pd
 from aequilibrae.project.database_connection import database_connection
 
 from aequilibrae.transit import Transit
 from aequilibrae.utils.create_example import create_example
-# sphinx_gallery_thumbnail_path = 'images/plot_import_gtfs.png'
+
```
pedrocamargo, line 26: Why remove the thumbnail?

docs/source/modeling_with_aequilibrae/static_traffic_assignment/assignment_mechanics.rst
```diff
@@ -325,6 +325,36 @@ There are still three parameters missing for the assignment.
   assig.set_time_field("free_flow_time")
   assig.set_algorithm(algorithm)
 
+
+Setting Preloads
+^^^^^^^^^^^^^^^^
+
+We can also optionally include a preload vector for constant flows which are not
+being otherwise modelled. For example, this accounts for predetermined public transport 
+vehicles at their regularly scheduled times, and reduces the capacity on each link
```
pedrocamargo: It does not reduce capacity

tests/aequilibrae/transit/rebuild_transit_example.ipynb
```diff
@@ -0,0 +1,137 @@
+{
```
pedrocamargo: Are you sure this notebook s part of the PR?

pedrocamargo:
@mitsudz , missing some linting as well.

tests/aequilibrae/transit/rebuild_transit_example.ipynb
mitsudz: I was asked to put this in the repo as an example - see AequilibraE Teams chat under General.

docs/source/examples/creating_models/plot_import_gtfs.py
mitsudz: Accidental, reverted changes to this file now

docs/source/modeling_with_aequilibrae/static_traffic_assignment/assignment_mechanics.rst
mitsudz: Updated

pedrocamargo (approved):

pedrocamargo:
@Art-Ev , your pre-loading is ready to merge!!!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 120
-target-version = ['py39', 'py310', 'py311', 'py312', 'py313']
```
jamiecook, line 3: py313 did not work for me - thoughts @pedrocamargo

pedrocamargo: It is a Black version thing, I believe. It works well for me.

https://black.readthedocs.io/en/stable/change_log.html

Art-Ev:
Hi ! Could the preloading function be used the same way as capacity or free_flow_time definition functions ? One “set_preload_field” function which, if not set/empty, does not apply preloading for example ? ( would be easier to use than create by yourself a dedicated dataframe, which is ok... but not as smooth ), @pedrocamargo or @mitsudz ?

pedrocamargo:
The idea is that the user would be allowed to use whatever preloading they would like. Ther is a method for Transit that computes the preloading in the correct format for whatever time-period the user chooses, which they can then inject into the assignment. Does that make sense?

PR #529, opened 2024-05-27
Refactors transit map-matching

pedrocamargo:
This PR fixes the map-matching algorithm to prevent repeated links from showing repeatedly and to ensure the entire path is represented.  It also drastically reduces memory use by removing the PathStorage feature in favor of the more recent early-exit feature on one-to-one path computation

- [x] Removes path storage feature in favour of computing paths with early-exit
- [x] Fixes map-matching


aequilibrae/transit/lib_gtfs.py
```diff
@@ -71,8 +56,7 @@ def __init__(self, network, agency_identifier, file_path, day="", description=""
         self.__do_execute_map_matching = False
         self.__target_date__ = None
         self.__outside_zones = 0
-        self.__has_taz = 1 if len(self.geotool.zoning.all_zones()) > 0 else 0
-        self.path_store = PathStorage()
+        self.__has_taz = 1 if len(self.project.zoning.all_zones()) > 0 else 0
```
jamiecook:
```suggestion
        self.__has_taz = len(self.project.zoning.all_zones()) > 0
```

r-akemii (approved):

aequilibrae/transit/transit_elements/pattern.py
```diff
@@ -26,9 +27,9 @@ class Pattern(BasicPTElement):
     def __init__(self, route_id, gtfs_feed) -> None:
         """
         :Arguments:
-            *pattern_id* (:obj:`str`): Pre-computed ID for this pattern
+            *route_id* (:obj:`str`): route ID for which this stop pattern belongs
 
-            *geotool* (:obj:`Geo`): Suite of geographic utilities.
+            *gtfs_feed* (:obj:`Geo`): Parent feed object
```
jamiecook, line 32: I don't think that's a Geo object type.

PR #531, opened 2024-06-02
Route choice

pedrocamargo:
To be merged when all issues solved


Jake-Moss:
@janzill This should now be fixed https://github.com/AequilibraE/aequilibrae/pull/532#issuecomment-2144114087. We're also now returning a column `mask` to indicate if the route was excluded because of the binary logit, not just setting it's probability to 0.

I plan to fix the disconnected ODs issue tomorrow morning. I think I have an idea of what is going on (and may have already partially fixed it)

janzill:
> @janzill This should now be fixed [#532 (comment)](https://github.com/AequilibraE/aequilibrae/pull/532#issuecomment-2144114087). We're also now returning a column `mask` to indicate if the route was excluded because of the binary logit, not just setting it's probability to 0.
> 
> I plan to fix the disconnected ODs issue tomorrow morning. I think I have an idea of what is going on (and may have already partially fixed it)

I tested this, it gives the correct results for path overlap, thanks Jake!

pedrocamargo:
Feel free to merge if it all looks good, @janzill

janzill (approved):
I've reviewed most of this as part of the route choice docs branch that was just merged, this looks good. @Jake-Moss just double-checking - do we require pyarrow==15.0.0?

Jake-Moss:
> I've reviewed most of this as part of the route choice docs branch that was just merged, this looks good. @Jake-Moss just double-checking - do we require pyarrow==15.0.0?

It's less of an actual dependency and more of a packaging issue. To get the wheel repair working on linux I had to specify the version of pyarrow. I'll give it another shot today after @pedrocamargo solved the macos build issues.

I've opened a discussion on the CIBuildWheels page asking for some general advice regarding this https://github.com/pypa/cibuildwheel/discussions/1872

Jake-Moss:
I'm very unsure how this packaging should be handled. As I understand it the repair process involves finding external libs and placing them inside the wheel and correcting the paths. I've got a couple concerns with this,
- If we continue to ignore pyarrow when repairing then we must depend on an explicit version of pyarrow, we're asking users to install the exact version of pyarrow that we built against.
- If we repair the wheel and place the `libarrow.so` and `libpython.so` inside the aequilibrae wheel them we're essentially repackaging pyarrow. This solves the version dependencies issues, however it will increase our wheel size. I think this is the "best" solution in terms of least issues. However, I haven't been able to get it to work, `auditwheel` and `delocate-wheel` are unable find the `libarrow.so` and`libpython.so` libraries. I think this is because the repair process doesn't have pyarrow install? But I've installed it and it's still been unable to locate it

janzill:
Thanks @Jake-Moss in that case I am happy for this to be merged

requirements.txt
```diff
@@ -8,4 +8,5 @@ pandas
 pyproj
 rtree
 openmatrix
-geopandas
\ No newline at end of file
+pyarrow==16.1.0
```
jamiecook, line 11: do we need an exact version here?

Jake-Moss: Yes unfortunately because we need to know the exact version of the dynamic library to remove when repairing the wheel here
```toml
# pyproject.toml
[tool.cibuildwheel.linux]
archs = ["auto", "aarch64"]
build = ["cp310-*", "cp311-*", "cp312-*"]
repair-wheel-command = [
  "auditwheel repair -w {dest_dir} {wheel} --exclude libarrow.so.1601 --exclude libarrow_python.so"
]
```

It's something that I'd like to fix but I'm unsure how. I've opened a discussion here asking for general advice on how we should do this https://github.com/pypa/cibuildwheel/discussions/1872

PR #532, opened 2024-06-02
Route choice docs and adjusting

pedrocamargo:
Adds route choice capabilities to AequilibraE. Items still missing:

- [x] Include treatment of cases where the network is disconnected for a given OD pair
- [x] Adding documentation for choice set generation (modeling_with_aequilibrae/route_choice/choice_set_generation.rst)
- [x] Adding example for the creation of route choice sets
- [x] Adding documentation for route choice (modeling_with_aequilibrae/route_choice/route_choice_model.rst)
- [x] Add a map with assignment results to the route choice example
- [x] Make the cutoff probability intuitive (I would expect a value of 0.01 to exclude all routes that have a probability os less than 1%)
- [x] Update API for route choice (prepare graph with utility and provide only path overlap parameter)


docs/source/examples/assignment_workflows/plot_route_choice.py
```diff
@@ -51,7 +64,8 @@
 project.network.graphs.keys()
 
 # let's say we want to minimize the distance
-graph.set_graph("distance")
+graph.network = graph.network.assign(utility=graph.network.distance * theta)
```
pedrocamargo, line 80: @janzill , this is what we were missing.

@Jake-Moss , we could have a method in the graph object to allow for a dictionary to be provided while we do the math in the background, but I don't think that's needed

docs/source/modeling_with_aequilibrae/route_choice/choice_set_generation.rst
```diff
@@ -0,0 +1,124 @@
...
+recommended by Rieser-Schüssler et al. [1]_, which is also a reference for further details on the BFS-LE algorithm.
+
+Experiment
+----------
+
+In an experiment of with nearly 9,000 observed vehicle GPS routes covering a large Australian State, we found that
+all three algorithms (LP, BFS-LE, and BFS-LE+LP) had excellent performance in reproducing the observed routes. However,
+the computational overhead of BFS-LE is substantial enough to recommend always verifying if LP is fit-for-purpose.
+
+.. image:: ../../images/route_choice_sets.png
```
pedrocamargo, line 78: @Jake-Moss , maybe change this picture to include all three (LP, BFS-LE, BFS-LE+LP at 1%)

janzill:
@Jake-Moss it looks like we calculate path overlap factors before excluding paths with the provided minimum probability parameter. However, we want to exclude these paths from the choice set, which means we want to calculate path overlap factors AFTER exclusion.

janzill:
We might also want to think about using the minimum probability as an early stopping criterion for path searches.

docs/source/examples/assignment_workflows/plot_route_choice.py
```diff
@@ -119,14 +142,53 @@
...
+
+    long, lat = project.conn.execute("select avg(xmin), avg(ymin) from idx_links_geometry").fetchone()
+
+    map_osm = folium.Map(location=[lat, long], tiles="Cartodb Positron", zoom_start=12)
+    loads_lyr.add_to(map_osm)
+    folium.LayerControl().add_to(map_osm)
+    return map_osm
+
+# %%
+plot_results(rc.get_load_results()[0])
```
pedrocamargo: @Jake-Moss , why are we returning both compressed and uncompressed link loads?  The compressed ones don;t make sense for the user. Do we need it for anything?

docs/source/examples/assignment_workflows/plot_route_choice.py
```diff
@@ -119,14 +142,53 @@
...
+
```
pedrocamargo, line 206: @Jake-Moss , the results for the assignment don't make sense. at all. I am either making a silly mistake, or there is something very wrong indeed.
image

pedrocamargo: https://aequilibrae.com/python/dev/_auto_examples/assignment_workflows/plot_route_choice.html#sphx-glr-auto-examples-assignment-workflows-plot-route-choice-py

Jake-Moss:
> @Jake-Moss it looks like we calculate path overlap factors before excluding paths with the provided minimum probability parameter. However, we want to exclude these paths from the choice set, which means we want to calculate path overlap factors AFTER exclusion.

Ah damn, we are too. Shouldn't be a difficult fix

docs/source/examples/assignment_workflows/plot_route_choice.py
Jake-Moss: > @Jake-Moss , we could have a method in the graph object to allow for a dictionary to be provided while we do the math in the background, but I don't think that's needed

Not 100% sure what you mean here sorry? A dictionary to do what? Describe some function between columns?

docs/source/examples/assignment_workflows/plot_route_choice.py
Jake-Moss: We need it internally but otherwise no, I was just returning it because we computed it

pedrocamargo: Let's remove it, in that case.

docs/source/examples/assignment_workflows/plot_route_choice.py
pedrocamargo: Utility is not something we would have in the graph, so we have this little sequence  we have to do to of assigning a new field to the network and re-prepare the graph.  This could be done internally with a method (passing a dictionary with the function to be applied), but I don;t think that's needed.  

The important point is that the only parameter we should be passing to the PSL is the path-overlap factor (beta, I think you called in the code)

docs/source/examples/assignment_workflows/plot_route_choice.py
```diff
@@ -37,6 +43,19 @@
...
 
+# %%
+# Model parameters
+# ~~~~~~~~~~~~~~~~
+# We'll set the parameters for our route choice model. These are the parameters that will be used to calculate the
+# utility of each path. In our example, the utility is equal to *theta* * distance
+# And the path overlap factor (PSL) is equal to *beta*.
+
+# Distance factor
+theta = 0.00011
```
janzill, line 55: probably want this value to be negative

docs/source/examples/assignment_workflows/plot_route_choice_set.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+# %%
+# We grab the graph for cars
+graph = project.network.graphs["c"]
+
+# we also see what graphs are available
+project.network.graphs.keys()
+
+graph.set_graph("distance")
```
janzill, line 52: Are we not setting the utility here?

docs/source/modeling_with_aequilibrae/route_choice.rst
```diff
@@ -0,0 +1,31 @@
...
+Route Choice
+============
+
+AequilibraE is the first modeling package with full support for route choice, from the creation of choice sets through
+multiple algorithms to the assignment of trips to the network using the traditional Path-Size logit.
+
+Full process overview
+---------------------
+
+The estimation of route choice models based on vehicle GPS data can be explored on a family of papers scheduled to
```
janzill, line 67: technically none of these have been officially accepted yet

docs/source/modeling_with_aequilibrae/route_choice/choice_set_generation.rst
```diff
@@ -0,0 +1,125 @@
...
+Further research could explore the use of specialised integer vector hash functions. As we did not find the hashing
+had a non-negligible influence on the runtime performance, this optimisation was not tested.
+
+AequilibraE also implements a combination of LP and BFS-LP as an optional feature to the latter algorithm, as
+recommended by Rieser-Schüssler et al. [1]_, which is also a reference for further details on the BFS-LE algorithm.
+
+Experiment
+----------
+
+In an experiment of with nearly 9,000 observed vehicle GPS routes covering a large Australian State, we found that
```
janzill: remove 4th word (of)

docs/source/modeling_with_aequilibrae/route_choice/choice_set_generation.rst
```diff
@@ -0,0 +1,125 @@
...
+   graph.prepare_graph(np.array([1, 2, 3, 50, 100, 150))
+
+
+   nodes = [(1, 50), (2, 100), (3, 150)]  # List of tuples with (origin, destination) nodes
+   max_routes = 10  # Maximum number of routes to be computed for each OD pair
+   penalty = 1.01  # Penalty to be applied to links used in paths.
+   cores = 60  # Number of cores to be used in the computation
+   psl = True  # If True, the path size logit will be used to compute probabilities already
+   bfsle = True # Should we use BFSLE? If False, defaults to Link Penalization
+   # This is only useful if you are already using an utility measure to compute paths
```
janzill, line 109: unclear comment

docs/source/modeling_with_aequilibrae/route_choice/route_choice_model.rst
```diff
@@ -0,0 +1,124 @@
...
+   from aequilibrae import Project
+
+   proj = Project()
+   proj.load('path/to/project/folder')
+
+   proj.network.build_graphs()
+   graph = proj.network.graphs['c']
+
+   # Measure that will be used to compute paths
+   graph.network = graph.network.assign(utility=graph.network.distance * theta)
```
janzill: define theta

docs/source/modeling_with_aequilibrae/route_choice/route_choice_model.rst
```diff
@@ -0,0 +1,124 @@
...
+principles with certain assumptions on the distribution of the random part of the utility. To
+account for the correlation of alternatives, Ramming (Ramming, 2002) introduced a correction factor that measures the
+overlap of each route with all other routes in a choice set based on shared link attributes, which gives rise to the PSL
+model. The PSL model’s utility function is defined by
+
+.. math:: U_i = V_i + \beta_{PSL} \cross \log{\gamma_i} + \varepsilon_i
+
+with path overlap correction factor
+
+.. math:: \gamma_i = \sum_{a \in A_i} \frac{l_a}{L_i} \cross \frac{1}{\sum_{k \in R} \delta_{a,k}}
```
janzill: our implementation does not use length, but cost (i.e., negative utility), right?

janzill: I think this is from our ATRF paper, for which I calculated the path overlap factors with length "by hand" (in my estimation notebooks). I checked the influence on estimation results and it didn't matter, but we should probably make sure the documentation is consistent with our implementation.

aequilibrae/paths/route_choice.py
```diff
@@ -75,7 +75,7 @@ def set_choice_set_generation(self, /, algorithm: str, **kwargs) -> None:
 
         Setting the parameters for the route choice:
 
-        `beta`, `theta`, and `seed` are BFSLE specific parameters.
+        `beta` and `seed` are BFSLE specific parameters.
```
janzill: `beta` is the utility parameter on path overlap for probability calculations when using path size logit

aequilibrae/paths/route_choice.py
```diff
@@ -104,8 +104,8 @@ def set_choice_set_generation(self, /, algorithm: str, **kwargs) -> None:
         excluded from the PSL calculations. The route is still returned, but with a probability of 0.0.
 
         The `cutoff_prob` should be in the range [0, 1]. It is then rescaled internally to [0.5, 1] as probabilities
```
janzill, line 106: Maybe we should clarify the comment regarding re-scaling or move it to the implementation below. The bottom line is that we are using a binary logit model between the shortest path and each path in the choice set, which means that a probability of 0.5 corresponds to equal utility. The shortest path will always have higher utility than all produced paths so  will always have probability larger than 0.5 and therefore providing larger values would lead to exclusion of all alternatives. We just re-scale to make the cutoff probability parameter easier to interpret.

aequilibrae/paths/route_choice_set.pyx
```diff
@@ -254,8 +257,7 @@ cdef class RouteChoiceSet:
             where: Optional[str] = None,
             path_size_logit: bool = False,
             beta: float = 1.0,
-            theta: float = 1.0,
-            cutoff_prob: float = 1.0,
+            cutoff_prob: float = 0.0,
     ):
         """Compute the a route set for a list of OD pairs.
```
janzill, line 263: can we add the missing arguments please

docs/source/modeling_with_aequilibrae/route_choice/choice_set_generation.rst
janzill: actually, should probably just go to psl one line up

janzill (approved):
a couple of minor comments but looking really good

docs/source/examples/assignment_workflows/plot_route_choice_set.py
janzill: ah we do not care about probabilities because we only want to plot the route set

docs/source/examples/assignment_workflows/plot_route_choice.py
pedrocamargo: The negative is in the Cython code...

docs/source/modeling_with_aequilibrae/route_choice.rst
pedrocamargo: I know... But I am sure they will be...

pedrocamargo:
@janzill , since we are saying that utility is positive and the negative sign is in the computation itself, then I think we should do three things:

* Make sure the math is correct also for the binary logit
* Document that assumption (needed to compute paths, actually)

aequilibrae/paths/route_choice_set.pyx
pedrocamargo: which missing arguments?

aequilibrae/paths/route_choice.py
pedrocamargo: Agreed. Can you make that change?

janzill:
> @janzill , since we are saying that utility is positive and the negative sign is in the computation itself, then I think we should do three things:
> 
> * Make sure the math is correct also for the binary logit
> * Document that assumption (needed to compute paths, actually)

Hm, maybe we should refer to cost instead of utility. The cost of a link is a measure that has to be positive for the shortest path algorithm to work, and we need our total cost to be the sum of link costs. The observed utility in the path size logit calculations is then the negative path cost plus parameter times path overlap term (which itself is calculated with link costs). I double-checked, the maths should be correct using cost.

aequilibrae/paths/route_choice.py
janzill: I left it here and just added a clarification.

janzill (approved):

PR #535, opened 2024-06-07
Adds support to multiple transit agencies

### LINKED ISSUE #528: Add support to more than one transit agency in GTFS feeds
Currently, AequilibraE only supports a single transit agency for each feed import. 

If our _agency.txt_ file has more than one agency, all of them are saved under the umbrella of the agency defined in the import 
`transit = data.new_gtfs_builder(agency="agency_name", file_path="gtfs_file_path")`.

It would be nice if AequilibraE supported more than one transit agency at a time, without the requirement of providing a transit agency name during the feed import.

r-akemii:
In this PR we add support to import more than one transit agency per GTFS file. We also modified how stops are numbered and its agency_id.

Modifications include:

* allows the import of multiple transit agencies per feed
* removes stops numbering based on transit agency
* updates column selection for GTFS files
* updates examples in docs (_plot_import_gtfs.py_ and _plot_public_transit_assignment.py_): removed agency argument
* updates link to GTFS documentation in _modeling_with_aequilibrae/public_transport.rst_ and on database specifications (_aequilibrae/project/database_specification/transit/tables_)
* updates tests in paths, project, and transit


r-akemii:
An initial thought was import transit agency data first, and have a workflow similar to the one that exists in `__load_date` (transit/gtfs_loader.py). I had an initial thought of loading all data for one agency at a time (like load all data for agency A, then load all data for agency B), but I don't think it's a good idea.

aequilibrae/project/database_specification/transit/tables/fare_zones.sql
```diff
@@ -4,12 +4,8 @@
 --@ **fare_zone_id** identifies the fare zone for a stop
 --@ 
 --@ **transit_zone** identifies the TAZ for a fare zone
---@ 
---@ **agency_id** identifies the agency fot the specified route
 
 CREATE TABLE IF NOT EXISTS fare_zones (
-	fare_zone_id    INTEGER  PRIMARY KEY,
-	transit_zone    TEXT     NOT NULL,
-	agency_id       INTEGER  NOT NULL,
```
pedrocamargo: Why this? We CAN have different fare zones for different operators

aequilibrae/project/database_specification/transit/tables/stops.sql
```diff
@@ -32,17 +28,14 @@
 CREATE TABLE IF NOT EXISTS stops (
 	stop_id           TEXT     PRIMARY KEY,
 	stop              TEXT     NOT NULL ,
-	agency_id         INTEGER  NOT NULL,
```
pedrocamargo: It is better to keep the agency ID and use the agency ID with the most services going through that stop.

aequilibrae/project/database_specification/transit/tables/fare_zones.sql
pedrocamargo: Let's make that fare_zone_id is names transit_fare_zone
And that should be something like 100,000 * agency_id + actual fare zone (or some counter for it)

pedrocamargo: Let's allow for it to be empty in case there are multiple feeds inside that GTFS

PR #537, opened 2024-06-17
Adds google analytics setup

pedrocamargo:


docs/source/conf.py
```diff
@@ -137,6 +137,9 @@
     },
     # "check_switcher": False,
     "github_url": "https://github.com/AequilibraE/aequilibrae",
+    "analytics": {
+        "google_analytics_id": "G-0HRKZDXRZ7",
+    },
```
r-akemii, line 142: @pedrocamargo I set the configuration here as in the [analytics page](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/analytics.html) in the user guide. 

From what I looked into Pydata theme documentation, there's no need to use Google Analytics template nor the section with `html_js_files` and `html_context`, but I don't know how the analytics tool from this point forward.

PR #545, opened 2024-06-21
avoids division by zero in Cython

pedrocamargo:
@Jake-Moss ,  this one was failing when running assignment, so this fixes it. I am not sure the math is correct, though

Let's also include in this PR:

- [ ] Ability to assign matrices with type float32 in route choice
- [x] Ability to prepare RC assignment with the OD pairs for which there is flow only
- [x] Handling disconnected ODs

### DIFF
--- a/aequilibrae/paths/route_choice.py
+++ b/aequilibrae/paths/route_choice.py
@@ -21,14 +21,14 @@ from aequilibrae.paths.route_choice_set import RouteChoiceSet
 class RouteChoice:
     all_algorithms = ["bfsle", "lp", "link-penalisation", "link-penalization"]
 
-    default_paramaters = {
-        "generic": {"seed": 0, "max_routes": 0, "max_depth": 0, "max_misses": 100, "penalty": 1.01, "cutoff_prob": 0.0},
+    default_parameters = {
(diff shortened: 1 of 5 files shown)


aequilibrae/paths/route_choice.py
```diff
@@ -174,7 +182,21 @@ def prepare(self, nodes: Union[List[int], List[Tuple[int, int]]]) -> None:
...
         """
+        if nodes is None:
+            idx = self.matrix.index
+            if len(self.matrix.view_names) > 1:
+                m = self.matrix.matrix_view.sum(axis=2)
+            else:
+                m = self.matrix.matrix_view
+            non_zero = np.where(m > 0)
+            o = non_zero[0][non_zero[0] != non_zero[1]]
+            d = non_zero[1][non_zero[0] != non_zero[1]]
```
jamiecook, line 195: maybe worth a comment here
```suggestion
            # Remove intrazonals (elements on the diagonal)
            o = non_zero[0][non_zero[0] != non_zero[1]]
            d = non_zero[1][non_zero[0] != non_zero[1]]
```

Jake-Moss: Thanks, I've added it to my local copy on the other PR

PR #547, opened 2024-06-25
Route choice refactor

Jake-Moss:
This PR aims to refactor the route choice module into something sane. It's grown too big and minor changes have a larger than desired impact. The `RouteChoiceSet` class is concerned with too much and needs to be broken up. With this I hope we can,
- Inline link loading, this will allow us to significantly reduce memory usage when the route sets are part of the desired output.
- Support float32 and float64 matrix types.
- Make the module more extensible and robust, it's rather fragile and assumes a lot about the state of itself.

I plan to
- [x] break out results storage and computation into a separate class,
- [x] inline link loading,
- [x] parallelise link loading,
- [x] allow float32 and float64 matrix types
- [x] Add a check to ensure that there are no two matrices (columns) in the GeneralisedCOODemand with the same name

Supersedes #545


pedrocamargo:
@Jake-Moss , I messed up and commited straight into your branch, but there was nothing controversial in them, so I continued with it.

Also, can I ask you to break things into more files to make it easier to find things later?  Things like the generalized COO demand, for example.  We can put the Cython files into subfolder and compile them into modules in the current folder to make it cleaner, as Jamie did for the QFM.  No priority for that, of course.

Jake-Moss:
> @Jake-Moss , I messed up and commited straight into your branch, but there was nothing controversial in them, so I continued with it.

No worries, perfectly happy to have those commits here, I would have merged them immediately anyway.

> Also, can I ask you to break things into more files to make it easier to find things later? Things like the generalized COO demand, for example. We can put the Cython files into subfolder and compile them into modules in the current folder to make it cleaner, as Jamie did for the QFM. No priority for that, of course.

Of course, the file is certainly getting unwieldy.

Jake-Moss:
This PR now contains https://github.com/AequilibraE/aequilibrae/pull/541

I've added support for select link loading link loads, but not the OD matrix ATM. That will come tomorrow morning. The SL implementation appears to be working well from a quick sanity check in qgis, just needs to tests to be updated.

Jake-Moss:
This is now nearing completion, however I've been running into some crashes when mixing f32 and f64, as well as on large OD pair counts. I haven't been able to narrow these down but I'm sure it's just some miss-allocation/typo somewhere

pedrocamargo:
@Jake-Moss , all sorts of tests still failing, so let's wait a little to finish this up.

pedrocamargo (approved):

PR #556, opened 2024-08-06
need __version__.py if user needs to compile for their architecture or python version.

jamiecook:


pedrocamargo:
The paths for some of the pyx files changed (to ./cython), and that has to change in the setup.py as well. Something that @r-akemii can take a look at?

setup.py
```diff
@@ -100,7 +100,7 @@
     "aequilibrae.reference_files": ["spatialite.sqlite", "nauru.zip", "sioux_falls.zip", "coquimbo.zip"],
     "aequilibrae.paths": ["parameters.pxi", "*.pyx"],
```
pedrocamargo: "aequilibrae.paths.cython": ["*.pix", "*.pyx" "*.pxd"],

aequilibrae/__init__.py
```diff
@@ -11,7 +11,7 @@
 from aequilibrae import project
 
 try:
-    from aequilibrae.paths.AoN import path_computation
+    from aequilibrae.paths.cython.AoN import path_computation
```
pedrocamargo: Wrong. The extension is NOT built inside the cython folder

aequilibrae/paths/__init__.py
```diff
@@ -9,14 +9,14 @@
 from aequilibrae.paths.vdf import VDF
 from aequilibrae.paths.graph import Graph, TransitGraph
 from aequilibrae.paths.optimal_strategies import OptimalStrategies
-from aequilibrae.paths.public_transport import HyperpathGenerating
+from aequilibrae.paths.cython.public_transport import HyperpathGenerating
```
pedrocamargo: again

pedrocamargo:
@Jake-Moss , can you cleanup this mess we made?  :-D

Jake-Moss:
Looks like I removed `__version__.py` from the package here https://github.com/AequilibraE/aequilibrae/commit/230d51ea04c4549a076e853a48b8136bc3bfc392 because of the request here https://github.com/conda-forge/staged-recipes/pull/22437#pullrequestreview-1980982844 . It was being packaged at the top level of wheel, which when unpacked would be outside of the `aequilibrae` package. The same was happening with our `test` dir but that should have been fixed via the `FlatLayoutPackageFinder`.

I'm not sure what's breaking because `setup.py` is not packaged in the wheel, which is the only use of `__version__.py` I can find.

PR #573, opened 2024-09-20
Significantly improve graph mapping on uncompressible networks (~8x)

Jake-Moss:


Jake-Moss:
This brings the QFM network map building down from 2m 40s to ~19s on my system. I think this could go further but might not be worth it.

This should produce exactly the same results as the previous version

aequilibrae/paths/cython/graph_building.pyx
```diff
@@ -442,15 +469,19 @@ def create_compressed_link_network_mapping(graph):
...
 
-        graph.compressed_link_network_mapping_idx = np.array(idx)
-        graph.compressed_link_network_mapping_data = np.array(data)
-        graph.network_compressed_node_mapping = np.array(node_mapping)
+    graph.compressed_link_network_mapping_idx = np.array(idx)
+    graph.compressed_link_network_mapping_data = np.array(data)
+    graph.network_compressed_node_mapping = np.array(node_mapping)
 
-        return idx, data, node_mapping
+    print("end compressed link network mapping", time.time() - start_tiem)
```
pedrocamargo: @Jake-Moss , let's remove this printing.  MMaybe put as a logging.debug?

aequilibrae/paths/cython/graph_building.pyx
```diff
@@ -387,50 +388,76 @@ def build_compressed_graph(graph):
...
+        long long i, j, a_node, x, b_node, tmp, compressed_id, dup_idx
+        long long[:] b
+        long long[:] values
+        np.uint32_t[:] idx
+        np.uint32_t[:] data
+        np.int32_t[:] node_mapping
+        np.int64_t[:, :] dups
+
+    start_tiem = time.time()
+    print("starting compressed link network mapping", start_tiem)
```
pedrocamargo: This one goes

aequilibrae/paths/cython/graph_building.pyx
Jake-Moss: Whoops I have a commit for that just mustn't have pushed it

pedrocamargo (approved):

PR #579, opened 2024-10-17
Connectivity tester

pedrocamargo:
Adds a connectivity analysis checker based on Depth-First search.  Requires very little memory to run.


aequilibrae/paths/connectivity_analysis.py
```diff
@@ -0,0 +1,72 @@
+import multiprocessing as mp
+import sys
+
+from aequilibrae.paths.AoN import connectivity_multi_threaded
+
+from aequilibrae.utils.core_setter import set_cores
+from aequilibrae.utils.signal import SIGNAL
+
+sys.dont_write_bytecode = True
```
Jake-Moss, line 9: What does this do? I can actually see 3 of these in the repo
```
Exported grep results:

aequilibrae/paths/network_skimming.py:16:sys.dont_write_bytecode = True
aequilibrae/paths/connectivity_analysis.py:9:sys.dont_write_bytecode = True
aequilibrae/__init__.py:38:    sys.dont_write_bytecode = True
```

Jake-Moss (approved):
Looks good to me

aequilibrae/paths/connectivity_analysis.py
Jake-Moss: I'm going to resolve this to merge the PR as I don't think this is an issue, more just curious

PR #580, opened 2024-10-18
Fix possible memory error in IPF, use Cython fused type

Jake-Moss:
This fixes a possible memory error in IPF where another thread may overwrite part of the work that's already been completed. It also uses a Cython fusedtype `cython.floating` to reduce code duplication

### DIFF
--- a/aequilibrae/distribution/cython/ipf_core.pyx
+++ b/aequilibrae/distribution/cython/ipf_core.pyx
@@ -4,7 +4,8 @@ import multiprocessing as mp
 cimport cython
 import numpy as np
 from cython.parallel import parallel, prange
-from libc.stdlib cimport malloc, free
+from libc.stdlib cimport calloc, free
+from libc.string cimport memset
 
 
 def ipf_core(seed_matrix, target_productions, target_attractions, max_iterations=200, tolerance=0.001, cores = 0, warn=True):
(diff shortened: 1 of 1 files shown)


pedrocamargo (approved):

aequilibrae/distribution/cython/ipf_core.pyx
```diff
@@ -32,44 +33,34 @@ def ipf_core(seed_matrix, target_productions, target_attractions, max_iterations
     cdef double [:] prod_tgt = target_productions
     cdef double [:] prod_factor = factor_prod
 
-    cdef double [:] attr_tot = mat_attr_tot
+    cdef double [::1] attr_tot = np.ascontiguousarray(mat_attr_tot)
```
jamiecook, line 36: hey @Jake-Moss  - what is the significance/meaning of the `::1` here?

Jake-Moss: That is to declare the memory view as contiguous so we can safely memset it later. Ideally this should be done whenever a memory view and C functions interact but it's fine 99% of the time in our usage because we aren't passing in strided numpy arrays

https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#pass-data-from-a-c-function-via-pointer
https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#declaring-the-numpy-arrays-as-contiguous

PR #581, opened 2024-10-18
Set based early exit

Jake-Moss:
This adds the basis for having set based early exit. It's currently not functional, I'm sure I've mistaken something for something else somewhere


Jake-Moss:
CI passed before I pushed that last commit

pedrocamargo (approved):

PR #583, opened 2024-10-25
Deprecates AequilibraEData in favor of Pandas DataFrames

pedrocamargo:


Jake-Moss:
Just trying this out locally now, there's a couple failures but nothing severe. I do think we should sort out the pandas specific warnings, I got about 40 while running the tests, ~12  of those are from pandas

aequilibrae/matrix/aequilibrae_matrix.py
```diff
@@ -283,8 +283,8 @@ def get_matrix(self, core: str, copy=False) -> np.ndarray:
 
     def create_from_omx(
         self,
-        file_path: str,
         omx_path: str,
+        file_path: str = None,
```
Jake-Moss, line 287: This is actually a breaking change, people who relied on the position of the arguments (us, in our docs) will have the file paths swapped

Jake-Moss (approved):
Looks like CI is happy with me now. Changes look good to me

aequilibrae/matrix/aequilibrae_matrix.py
pedrocamargo: True, but there is not such way out, as the best behaviour is to have it as a memory-only matrix.

I would doubt anyone uses this, however.

PR #586, opened 2024-11-07
Fixes check when creating links table

r-akemii:
In this PR we modify the direction check when creating "links" table. 

Previously, checking the length raised a `sqlite3.IntegrityError`  when updating/creating a link with direction `-1`.

We also add a test to check if the conditions are met and errors are raised when setting the direction values equal to -2, -1, 0, 1, 2.


jamiecook (approved):
Looks good to me, and tested too - lgtm

PR #589, opened 2024-11-15
Revert "Use default wheel repair (#585)"

Jake-Moss:
This reverts commit 659b7894825c176021fb517a65049c0b2aa20f13.

I'm unable to make this work. Because we are linking against the library from pyarrow installed as a dependency, once the wheel is build the temporary environment cibuildwheels constructs to build the wheel is destroyed so we cannot access it to properly repair the wheel. From what I can tell ignoring it is the only way.

This was intended as fix for the conda forge build, but it looks like I'll have to find some fix for it on that end

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ max-complexity = 20
 
 [tool.cibuildwheel]
 # While we test this facility we we will only build for 3.10
+#build = ["cp310-*"]
 build = ["cp39-*","cp310-*", "cp311-*", "cp312-*"]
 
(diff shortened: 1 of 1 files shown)


Jake-Moss:
I thought I'd found library that was doing a similar thing to us (https://github.com/blue-yonder/turbodbc) however, they don't ship any wheels, only source distributions and conda forge. This side steps all the issues we've had with wheels and pyarrow.

pedrocamargo (approved):

PR #590, opened 2024-11-15
Better load balancing for route choice

pedrocamargo:
When running Route choice on VERY LARGE networks, load balancing can be pretty poor with fixed chunk sizes.


Jake-Moss (approved):
I haven't seen that option before but it seems quite reasonable to me.

> As with dynamic scheduling, the iterations are distributed to threads as they request them, but with decreasing chunk size. The size of each chunk is proportional to the number of unassigned iterations divided by the number of participating threads, decreasing to 1 (or the chunksize if provided).
This has an advantage over pure dynamic scheduling when it turns out that the last chunks take more time than expected or are otherwise being badly scheduled, so that most threads start running idle while the last chunks are being worked on by only a smaller number of threads.

PR #592, opened 2024-11-20
Fails hard on import

pedrocamargo:
This would make errors more obvious.  

It may be a little bit of a pain when dealing with non-binary-dependent procedures, but worth the change


Jake-Moss (approved):

jamiecook:
Something to consider is whether we can make the error message around 'no-compiled binaries' a bit easier to understand - I've seen a few users posting confused why it can't import AoN, there's nothing at the moment to point them in the right direction.

Issue #591, opened 2024-11-19
DLL load failed while importing AoN in Aequilibrae


zahern:
Description:
I encountered an ImportError while attempting to run a script that imports the Project class from the aequilibrae package. The traceback indicates that the error occurs when trying to import the AoN module from the aequilibrae.paths package.

I am running Python 3.10 on windows:

Exact error gives :

  File "...\.venv\lib\site-packages\aequilibrae\paths\__init__.py", line 1, in <module>
    from aequilibrae.paths.AoN import one_to_all, skimming_single_origin, path_computation, update_path_trace
ImportError: DLL load failed while importing AoN: The specified module could not be found.

pedrocamargo:
Can you provide more information on the environment you are running the code on?  With the above information is not possible to reproduce the problem.

zahern:
My Python version is 3.10.11.
Running a virtual env where:
aequilibrae is 1.1.3.


Looking at the Package contents into \aequilibrae\paths\ I have the following files listed in the screenshot
<img width="1080" alt="Equilibrae_python" src="https://github.com/user-attachments/assets/3549dd1b-35fe-4a16-8fed-265936dea8d2">
I believe AoN is not builiding properly as my file name in the list is called "AoN.cp310-win_amd64.pyd"

Am I doing anything wrong?

pedrocamargo:
This is still insufficient to reproduce the issue.  

It might have something to do with the version of PyArrow or NumPy you have installed, but it is impossible to guess

pedrocamargo:
You can try running the line below to see if you obtain a more informative error message:

from aequilibrae.paths.AoN import one_to_all, assign_link_loads

zahern:
Sorry Pedro, I have attached my installed packages.
[requirements.txt](https://github.com/user-attachments/files/17823156/requirements.txt)


I have tried running only "from aequilibrae.paths.AoN import one_to_all, assign_link_loads"

And i receive this error 
<img width="1159" alt="import" src="https://github.com/user-attachments/assets/af933fdd-2c54-4758-96fb-29fe4a2e5359">

pedrocamargo:
The issue turned out to be elsewhere in the compilation, and only affected people who did not have a development setup on their Windows environments.  This was fixed on #601

PR #601, opened 2025-01-15
Replace openmp:llvm with normal openmp, fix signed-ness errors

Jake-Moss:
In my dev VM this removed the linkage against `libomp`. Hopefully this resolves the import issues.

Thanks to @jamiecook for finding the relevant docs
> Starting in Visual Studio 2019 version 16.9, you can use the experimental /openmp:llvm option instead of /openmp to target the LLVM OpenMP runtime. **Support currently isn't available for production code, since the required libomp DLLs aren't redistributable**. The option supports the same OpenMP 2.0 directives as /openmp. And, it supports all the SIMD directives supported by the /openmp:experimental option. **It also supports unsigned integer indices in parallel for loops according to the OpenMP 3.0 standard**. For more information, see [Improved OpenMP Support for C++ in Visual Studio](https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/).
>
> https://learn.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-170

### DIFF
--- a/aequilibrae/paths/cython/route_choice_set.pyx
+++ b/aequilibrae/paths/cython/route_choice_set.pyx
@@ -199,7 +199,7 @@ cdef class RouteChoiceSet:
         """
         cdef:
             long long origin, dest
-            size_t i
+            long int i
 
         if select_links is None:
             select_links = {}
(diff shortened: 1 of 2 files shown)


pedrocamargo (approved):

pedrocamargo:
close #591

PR #603, opened 2025-01-22
fix int64 pandas conversion in transit graph builder

germanova:
On Windows, probably because of the pandas or numpy version, a conversion to int64 is required to prevent auto conversion to int32. This changes the travel time values for some edges.

### DIFF
--- a/aequilibrae/transit/transit_graph_builder.py
+++ b/aequilibrae/transit/transit_graph_builder.py
@@ -341,7 +341,7 @@ class TransitGraphBuilder:
         tt.dropna(how="any", inplace=True)
         tt[["last_departure", "last_trip_id", "last_pattern_id"]] = tt[
             ["last_departure", "last_trip_id", "last_pattern_id"]
-        ].astype(int)
+        ].astype(np.uint64)
 
         tt.loc[(tt.last_pattern_id != tt.pattern_id) | (tt.last_trip_id != tt.trip_id), "trav_time"] = np.nan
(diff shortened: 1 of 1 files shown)


Jake-Moss:
We were seeing `last_trip_id` and `trip_id` values going from `10001001001` (in decimal, int64 required) to `1411066409`. In binary this looks like truncation to 32 bits
```
before: 0010 0101 0100 0001 1011 0010 1010 0010 1001
after:       0101 0100 0001 1011 0010 1010 0010 1001
```

This was causing the trip id's to no long match and be dropped later. This gave a bunch of edges the max travel time. Because centroid flow must be enabled for this network we didn't see any issues with a route no longer existing due to the centroid connector generator method (overlapping regions).

aequilibrae/transit/transit_graph_builder.py
```diff
@@ -341,7 +341,7 @@ def _compute_segment_travel_time(self, time_filter=True):
         tt.dropna(how="any", inplace=True)
         tt[["last_departure", "last_trip_id", "last_pattern_id"]] = tt[
             ["last_departure", "last_trip_id", "last_pattern_id"]
-        ].astype(int)
+        ].astype(np.uint64)
```
pedrocamargo, line 344: Is this uint64 or np.int64?  Does it matter here?

pedrocamargo (approved):

pedrocamargo (approved):

PR #605, opened 2025-01-24
Fix transit_fare_zone values

### LINKED ISSUE #604: Wrong number of bindings on GTFS import for route systems with transit zones
GTFS for transit systems with fares zones, the import fails due to the wrong number of bindings in the SQL Query

File in question is: aequilibrae\aequilibrae\transit\lib_gtfs.py
Errors are on lines 247-250

Image

r-akemii:
This PR closes #604.

In this PR we:

* Fixed initial GTFS data wrangling for stops and fare_rules
* Fixed data saved into public_transport.database

For local testing purposes, we used this sample from Toronto (Canada): 
[sample_toronto.zip](https://github.com/user-attachments/files/18539604/sample_toronto.zip)

### DIFF
--- a/aequilibrae/transit/gtfs_loader.py
+++ b/aequilibrae/transit/gtfs_loader.py
@@ -224,8 +224,6 @@ class GTFSReader(WorkerThread):
             farerl = parse_csv(file, column_order[farerltxt])
         self.data_arrays[farerltxt] = farerl
 
-        corresp = {}
-        zone_id = self.agency.agency_id * AGENCY_MULTIPLIER + 1
         for line in range(farerl.shape[0]):
             data = tuple(farerl[line][list(column_order[farerltxt].keys())])
             fr = FareRule()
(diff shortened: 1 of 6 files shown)


aequilibrae/transit/gtfs_loader.py
```diff
@@ -234,13 +232,9 @@ def __load_fare_data(self):
...
             fr.agency_id = self.agency.agency_id
-            for x in [fr.origin, fr.destination]:
-                if x not in corresp:
-                    corresp[x] = zone_id
-                    zone_id += 1
-            fr.origin_id = corresp[fr.origin]
-            fr.destination_id = corresp[fr.destination] if fr.destination == "" else fr.destination_id
-            self.fare_rules.append(fr) if fr.origin == "" else fr.origin_id
+            fr.origin_id = int(fr.origin) if len(fr.origin) > 0 else None
+            fr.destination_id = int(fr.destination) if len(fr.destination) > 0 else None
```
pedrocamargo: Are you sure the ID has to be numeric, @r-akemii ?

r-akemii: I just checked the GTFS documentation and you're right! I used the wrong data type for origin/destination ID. It should be a sequence of UTF-8 characters. I'll fix it soon.

image

image

pedrocamargo (approved):

PR #606, opened 2025-01-28
Speeding up CI

pedrocamargo:


.github/workflows/unit_tests.yml
```diff
@@ -61,7 +62,7 @@ jobs:
 
     - name: Compile library
       run: |
-        python setup.py build_ext --inplace
+        python setup.py build_ext -j 4 --inplace
 
     - name: Runs test
       run: python -m pytest
```
Jake-Moss, line 68: With `pytest-xdist` I think we also need to run the tests in parallel.
```suggestion
      run: python -m pytest -n auto
```

With this do we need to disable/reduce the parallelism in the tests themselves?

Jake-Moss (approved):
Other than that one comment this looks good. Having the aarch64 runners is quite the improvement

.github/workflows/unit_tests.yml
pedrocamargo: We would need to solve the dependency problem...

PR #607, opened 2025-01-28
Version number for next release

pedrocamargo:
@r-akemii , there are to issues in the documentation that could be solved here: 

* something wrong with the paths for the develop version on the documentation.
* CI is not copying the documentation to the history folder upon release


jamiecook:
@pedrocamargo are you saying that those issues need to be resolved (by @r-akemii) on this PR?

PR #614, opened 2025-02-12
Reduce early exit set to only non-zero demand, add warning for missing centroids

Jake-Moss:
This reduces the early exit set size from all centroids to only those with non-zero demand. 

If there is zero demand for a particular origin early exit is disabled. This shouldn't happen as this condition is covered up in `allOrNothing.execute` but it's easy to check here as well. I've also removed the `if destination_count == 0:` check within `path_finding` as I found when it's hit it slows the assignment to a halt as the threads would fight over the gil. Now the path finding will just exit very quickly.

I've also added a warning for when there are centroids not found within the graph. This is hit by a few of our networks (particularly Coquimbo, thus it's not an exception), but I think it's a reasonable warning and would have made debugging easier.


aequilibrae/paths/cython/AoN.pyx
```diff
@@ -100,9 +117,6 @@ def one_to_all(origin, matrix, graph, result, aux_result, curr_thread):
         link_list = aux_result.select_links[:, :]  # Read only, don't need to slice on curr_thread
```
pedrocamargo, line 117: Doesn't the _nnz_destinations_  need to be equal to all centroids if we are skimming?

pedrocamargo (approved):

aequilibrae/paths/cython/AoN.pyx
Jake-Moss: Good point, we can still early exit if skimming, we just require all centroids be hit. I believe that is what we had previously.

PR #616, opened 2025-02-14
Small performance improvements

Jake-Moss:
With the early exit changes, it's now more likely to be fast to load the network normally than with a cascade loading.

For functions which one single dimensional, we found performance improvements by making them single threaded, removing the thread over head and time spent waiting. When Cython 3.1 releases we can revert those changes in favour of using the `use_threads_if` argument to get the best of both.

Also move ruff variables from top level to under the `lint` prefix as per the deprecation notice.

### DIFF
--- a/aequilibrae/paths/cython/basic_path_finding.pyx
+++ b/aequilibrae/paths/cython/basic_path_finding.pyx
@@ -32,41 +32,20 @@ cpdef void network_loading(long classes,
     cdef long long i, j, predecessor, connector, node
     cdef long long zones = demand.shape[0]
     cdef long long N = node_load.shape[0]
-# Traditional loading, without cascading
-#    for i in range(zones):
-#        node = i
-#        predecessor = pred[node]
-#        connector = conn[node]
-#        while predecessor >= 0:
-#            for j in range(classes):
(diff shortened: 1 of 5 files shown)


aequilibrae/paths/cython/basic_path_finding.pyx
```diff
@@ -32,41 +32,20 @@ cpdef void network_loading(long classes,
...
-    # Loads the demand to the centroids
-    for j in range(classes):
-        for i in range(zones):
-            if not isnan(demand[i, j]):
-                node_load[i, j] = demand[i, j]
-
-    # Recursively cascades to the origin
-    for i in range(found, 0, -1):
-        node = reached_first[i]
+    cdef double value
```
pedrocamargo: Do we still need the _reached_first_ array and _found_ variable to be fed here?

pedrocamargo (approved):

aequilibrae/paths/cython/basic_path_finding.pyx
Jake-Moss: Nope, looks there's a handful of other as well. Just running tests locally now

PR #620, opened 2025-02-17
Add skimming on s&f algorithm (#617)

pedrocamargo:


pedrocamargo:
@Jake-Moss , can you review and/or just approve (in case you have already reviewed) the Cython/agorithmic portions of this PR?

Jake-Moss (approved):
I think this is in a merge-able state now. I've added some tests for the skimming as well as fixed some memory leaks and general bugs.

I'm still not pleased with the `_update_od_values` method but my quick attempt to replacing it didn't get far. We don't need to realise all pairs of permutations of centroids to be able to skim. This will be easier to fix with the sparse matrix changes.

Additionally in the assignment I would rather use proper infinite and zero values instead of a really big number and a really small number. Though that was not changed as part of this PR. The original paper makes mention of the required changes to the algorithm I wasn't able to get it working.

PR #621, opened 2025-02-17
Generalizes table loader

pedrocamargo:
Generalizes procedures so we can use them with the transit tables

### DIFF
--- a/aequilibrae/utils/db_utils.py
+++ b/aequilibrae/utils/db_utils.py
@@ -7,8 +7,6 @@ from typing import Union
 
 import pandas as pd
 
-from aequilibrae.utils.spatialite_utils import connect_spatialite
-
 
 def list_tables_in_db(conn: Connection):
     sql = "SELECT name FROM sqlite_master WHERE type ='table'"
@@ -35,6 +33,7 @@ class commit_and_close:
 
             **missing_ok** (:obj:`bool`): Boolean indicating that the db is not expected to exist yet
         """
+        from aequilibrae.utils.spatialite_utils import connect_spatialite
 
         def get_conn() -> sqlite3.Connection:
             if spatial:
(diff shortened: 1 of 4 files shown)


aequilibrae/utils/db_utils.py
```diff
@@ -32,9 +33,19 @@ def __init__(self, db: Union[str, Path, Connection], commit: bool = True, missin
 
             **missing_ok** (:obj:`bool`): Boolean indicating that the db is not expected to exist yet
         """
```
Jake-Moss, line 35:
```suggestion
        **spatial** (:obj:`bool`): Boolean indicating if the database connection should use Spatialite
        """
```

aequilibrae/utils/db_utils.py
```diff
@@ -32,9 +33,19 @@ def __init__(self, db: Union[str, Path, Connection], commit: bool = True, missin
...
+            if spatial:
+                if not isinstance(db, (str, PathLike)):
+                    raise Exception("You must provide a database path to connect to spatialite")
+                return connect_spatialite(db, missing_ok)
+            elif isinstance(db, (str, PathLike)):
+                return safe_connect(db, missing_ok)
+            else:
+                return db
+
+        self.conn = get_conn()
```
Jake-Moss:
```suggestion
        if spatial:
            if not isinstance(db, (str, PathLike)):
                raise Exception("You must provide a database path to connect to spatialite")
            self.conn = connect_spatialite(db, missing_ok)
        elif isinstance(db, (str, PathLike)):
            self.conn = safe_connect(db, missing_ok)
        else:
            self.conn = db
```

This is more personal preference but I thought the `get_conn` method was used else where until I saw it was a lambda

Jake-Moss (approved):

PR #625, opened 2025-02-20
Speed up doc building

r-akemii:
This PR aims to reduce the total time to build AequilibraE's docs.

We:

* Removed unused menu in _navigation_header.html_
* Modified configuration in _conf.py_ (https://stackoverflow.com/questions/51986357/how-can-i-make-sphinx-build-build-faster)
* Modified workflow in _documentation.yml_ to avoid rebuilding the documentation and using XELatex engine, which is faster than latexpdf (https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.latex.LaTeXBuilder)
* Modified some plot examples to use `GeoDataFrame.explore()` with projects links and nodes rather than using folium with pandas iterrows


docs/zip_html.py
```diff
@@ -0,0 +1,24 @@
+import os
```
jamiecook: this script doesn't seem to be used... and would probably be better off as a call to `zip` or `tar` anyway.

.github/workflows/documentation.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 1
+      max-parallel: 20
```
jamiecook: does this do anything - reading the docs here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

I would expect this to only help if we had a matrix of jobs to run - multiple could be run at once... but we only have one documentation build i think...

pedrocamargo (approved):

PR #626, opened 2025-02-24
Efficient Centroid and centroid connector builder

pedrocamargo:


pedrocamargo:
@r-akemii , an example for the gallery would be ideal here

pedrocamargo:
@Jake-Moss , can you review, make minor corrections wherever needed and merge?

docs/source/examples/network_manipulation/plot_centroid_connection.py
```diff
@@ -0,0 +1,106 @@
...
+
+# %%
+# Then we connect the centroid to the network, by selecting the desired mode,
+# the number of connectors, the allowed link types for connection, and if one
+# wants to allow the connection to links in other zones. By setting ``limit_to_zone=False``,
+# we allow the centroid of one zone to be connected to a link outside the zone itself.
+
+zoning.connect_mode(mode_id=mode, connectors=1, limit_to_zone=False)
+
+# %%
```
Jake-Moss, line 79: When connecting multiple modes do we want to produce duplicate centroid connectors? When using this to connect the `b` and `c` modes, we end up with two connectors that are identical (other than IDs) 

image

Jake-Moss: It looks like the `connector_creation` function isn't aware of the a and b nodes it's connecting and just leaves it to db triggers to handle.

aequilibrae/project/network/connector_creation.py
```diff
@@ -19,6 +23,8 @@ def connector_creation(
     if len(mode_id) > 1:
         raise Exception("We can only add centroid connectors for one mode at a time")
 
+    warnings.filterwarnings("ignore", category=UserWarning, module="geopandas")
+
```
Jake-Moss: I'm not sure we want to ignore these warnings, this modifies the global state and it isn't undone. If we want to suppress a warning for a small part we should use the `warnings.catch_warnings` context manager to restore it after we are done.

Jake-Moss (approved):

docs/source/examples/network_manipulation/plot_centroid_connection.py
pedrocamargo: Can you address this, @r-akemii ?

aequilibrae/project/network/link.py
```diff
@@ -56,13 +56,13 @@ def delete(self):
             conn.execute(f'DELETE FROM links where link_id="{self.link_id}"')
         self.__stil_exists = False
 
-    def save(self):
+    def save(self, conn_=None):
```
pedrocamargo:
```suggestion
    def save(self, conn=None):
```

aequilibrae/project/network/link.py
```diff
@@ -56,13 +56,13 @@ def delete(self):
...
         """Saves link to database"""
 
         data, sql = self._save_new_with_geometry() if self.__new else self.__save_existing_link()
 
         if data:
-            with self.project.db_connection as conn:
+            with conn_ or self.project.db_connection as conn:
```
pedrocamargo:
```suggestion
            with conn or self.project.db_connection as conn:
```

aequilibrae/project/zone.py
```diff
@@ -121,6 +121,8 @@ def connect_mode(
             mode_id=mode_id,
             link_types=link_types,
             connectors=connectors,
+            proj_nodes=self.project.network.nodes.data,
```
pedrocamargo, line 124: This looks wrong. This basically loads ALL nodes each time a zone is instantiated

PR #630, opened 2025-03-03
Allows users to not have to download SpatiaLite on Windows

pedrocamargo:


aequilibrae/utils/spatialite_utils.py
```diff
@@ -38,7 +38,9 @@ def connect_spatialite(path_to_file: os.PathLike, missing_ok: bool = False) -> C
 
         return qgis.utils.spatialite_connect(path_to_file)
 
-    ensure_spatialite_binaries()
+    if os.environ.get("AEQ_ENSURE_SPATIALITE", "YES") != "NO":
```
Jake-Moss, line 41: I think this should be 
```suggestion
    if os.environ.get("AEQ_ENSURE_SPATIALITE", "YES") == "YES":
```
or
```suggestion
    if os.environ.get("AEQ_ENSURE_SPATIALITE", "NO") == "YES":
```

Otherwise things like `n`, `N`, `0`, and `no` will get interpreted as a "yes".

aequilibrae/utils/spatialite_utils.py
```diff
@@ -38,7 +38,9 @@ def connect_spatialite(path_to_file: os.PathLike, missing_ok: bool = False) -> C
...
+        ensure_spatialite_binaries()
+
```
Jake-Moss, line 43: Do we want the default to be an automatic download? I always prefer it when a program would ask first. In our case I sort of envision it as
- attempt to run spatialite query, query fails (catch this somehow?)
- raise an exception with a help message along the lines of "your missing spatialite, either download it yourself and place it in the SQLite load path, or use the `ensure_spatialite_binaries()` to automatically download it from <URL>, or set `AEQ_ENSURE_SPATIALITE=YES` to automaticity download it without asking"


This would be a little more difficult as we have a habit of opening a new sqlite connection pretty often (and closing them), rather than have one long lived connection and we create multiple cursors from it. If we just had one connection per db, then we would only need to load spatialite once per session.

aequilibrae/utils/spatialite_utils.py
```diff
@@ -66,11 +76,7 @@ def ensure_spatialite_binaries(directory: Optional[os.PathLike] = None) -> None:
     if not _dll_already_exists(directory):
         _download_and_extract_spatialite(directory)
 
-    # Update path and proj_lib env vars
-    if directory not in os.environ["PATH"]:
-        os.environ["PATH"] = directory + os.pathsep + os.environ["PATH"]
-    if "PROJ_LIB" not in os.environ:
-        os.environ["PROJ_LIB"] = directory
+    set_known_spatialite_folder(directory)
```
Jake-Moss, line 79: I didn't realise this was downloaded to a temp directory and not somewhere in site-packages. Can we download it once to avoid doing this every time?

Jake-Moss (approved):
I think the conditional should be changed to be very strict rather than accept anything that isn't `NO`, otherwise this looks good. It's useful to be able to configure this

aequilibrae/utils/spatialite_utils.py
pedrocamargo: The default is automatic download, which is today's behaviour

PR #641, opened 2025-04-03
Skimming post assignment

pedrocamargo:
Adds the capability of skimming post assignment (after convergence), so we can take advantage of matrix sparsity and early exit during assignment


aequilibrae/paths/graph.py
```diff
@@ -438,7 +438,13 @@ def set_skimming(self, skim_fields: list) -> None:
 
         if self.centroids.shape[0]:
             self.compact_skims = np.zeros((self.compact_num_links + 1, len(skim_fields) + 1), self.__float_type)
-            df = self.__graph_groupby.sum(numeric_only=True)[skim_fields].reset_index()
+
+            gpb = self.__graph_groupby
+            if any(x not in self.__graph_groupby for x in skim_fields):
+                gpb = self.graph.groupby(["__compressed_id__"])
+
```
Jake-Moss, line 445: What does this `if` do? `self.__graph_groupby` is set by `__build_compressed_graph` which we always run, so it's always set to `self.graph.groupby(["__compressed_id__"])`. But if none of the `skim_fields` are present in the groupby groups, then we set it again?

tests/aequilibrae/paths/test_traffic_assignment.py
```diff
@@ -47,6 +47,30 @@ def assignment(project):
...
 
 
+def test_skim_after(project, assigclass):
+    assig = TrafficAssignment(project)
+
+    assig.add_class(assigclass)
+    assig.set_vdf("BPR")
+    assig.set_vdf_parameters({"alpha": 0.15, "beta": 4.0})
+    assig.set_vdf_parameters({"alpha": "b", "beta": "power"})
+
```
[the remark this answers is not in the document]
Jake-Moss: Code is fine as is.

Jake-Moss (approved):

aequilibrae/paths/graph.py
pedrocamargo: The assignment costs and congested time are not part of the network object (they are directly on the graph and are temporary/transient fields), so they would not have been in the groupby we created from the beginning.

There is no setting it again because this is not something we compute and recompute during assignment

Jake-Moss: Perfect, thanks!

PR #644, opened 2025-04-16
Add model entry points

Jake-Moss:
This adds an  "entry point" module defined at `$project/entry_points/__init__.py`. Default parameters can be defined in `parameters.yml` under the `entry_points` header. The arguments are applied via `functools.partial` as keyword arguments and overwrite the original functions in the module. 

The module is dynamically loaded without modifying the python path and imported as a submodule of aequilibrae (`aequilibrae.entry_points`) to avoid potential name conflicts. I tried to get recursive module reloading (like `%autoreload`) working but with no luck. I think `importlib` and `IPython.lib.deepreload` really don't like this module.

I think just looking for `__init__.py` is simpler and more robust and reading individual files in.

ATM this unintentionally provides hot reloading of the `$project/entry_points/__init__.py` and `parameters.yml` files because we re-read/re-import them.

## TODO
- [x] Documentation
- [ ] Decide if we want to return the full module, or a `namedtuple` with just the functions mentioned in `parameters.yml`
- [x] Modify default examples/provide upgrade path for existing models
- [x] Write demo functions/example notebooks

## Notebook
<details>
<summary>Demo notebook</summary>

```python
# $project/entry_points/__init__.py
def func1(*args, **kwargs):
    print("func1")
    print(args)
    print(kwargs)

def func2(*args, **kwargs):
    print("func2")
    print(args)
    print(kwargs)
```

```yaml
# $project/parameters.yml
entry_points:
  func1:
    kwarg1: "something"
    kwarg2: 100
...
```

```python
import aequilibrae
from aequilibrae.utils.create_example import create_example
import tempfile
```

```python
file = tempfile.TemporaryDirectory(prefix="aequilibrae_")
project = create_example(file.name + "/project", "sioux_falls")
```

```python
project.project_base_path  # Need to create entry_points/__init__.py here 
project.project_parameters.restore_default()  # Restore defaults to override with new stuff
```

```python
module = project.entry_points
```

```python
module.func1(kwarg2="overridden")
```

> 
>     func1
>     ()
>     {'kwarg1': 'something', 'kwarg2': 'overridden'}
> 

```python
module.func2()
```

>     func2
>     ()
>     {}
> 

```python
# this is the actual module we imported, so we also get all the symbols created in that module. We just replace the symbols in it with partially applied functions
dir(module)
```

> 
>     ['__builtins__',
>      '__cached__',
>      '__doc__',
>      '__file__',
>      '__loader__',
>      '__name__',
>      '__package__',
>      '__path__',
>      '__spec__',
>      'func1',
>      'func2']

```python
# We can get the original function via the partial.func attribute
module.func1.func(kwarg2="overridden")
```

>     func1
>     ()
>     {'kwarg2': 'overridden'}

```python
# Or the applied arguments
module.func1.args, module.func1.keywords
```

>     ((), {'kwarg1': 'something', 'kwarg2': 100})

```python
# This is also a module you can just import now
import aequilibrae.entry_points as entry_points  # not sure if this is good behaviour, but it occurs since we need to add it to the sys.modules in order to call exec_module on it while importing
```

```python
entry_points.func1(kwarg2="overridden")
```

> 
>     func1
>     ()
>     {'kwarg1': 'something', 'kwarg2': 'overridden'}

</details>


aequilibrae/utils/model_run_utils.py
```diff
@@ -0,0 +1,31 @@
+import sys
+import pathlib
+import importlib.util
+from IPython.extensions import autoreload
```
pedrocamargo:
```suggestion
```

aequilibrae/parameters.yml
```diff
@@ -1,3 +1,7 @@
+entry_points:
```
pedrocamargo: This gets shipped with AequilibraE and is used when we create new projects, so it would make sense making it somewhat useful.  An idea is to have something that will create a series of models KPIs:

* List of matrices with their totals
* Number of Links, Nodes and Zones
* List of result tables in the model

pedrocamargo (approved):

PR #648, opened 2025-04-28
Run module example

r-akemii:
In this PR we format documentation strings and add an example in the gallery for AequilibraE's run module.


Jake-Moss (approved):
This looks good, thanks!

PR #656, opened 2025-05-13
Minor documentation changes

Jake-Moss:
- We shouldn't block centroid flow on Coquimbo, it has real centroids unlike Sioux falls
- Clarify that PT assignment only supports one matrix core at a time.

Thanks for finding these @r-akemii !

### DIFF
--- a/docs/source/examples/public_transport/plot_public_transit_assignment.py
+++ b/docs/source/examples/public_transport/plot_public_transit_assignment.py
@@ -77,7 +77,7 @@ periods.data
 project.network.build_graphs()
 
 # %%
-# Now we'll create the line strings for the access connectors, this step is optimal but provides more accurate distance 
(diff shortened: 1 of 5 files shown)


pedrocamargo (approved):

docs/source/examples/public_transport/plot_public_transit_assignment.py
```diff
@@ -78,7 +77,7 @@
 project.network.build_graphs()
 
 # %%
-# Now we'll create the line strings for the access connectors, this step is optinal but provides more accurate distance 
+# Now we'll create the line strings for the access connectors, this step is optimal but provides more accurate distance
```
pedrocamargo:
```suggestion
# Now we'll create the line strings for the access connectors, this step is optional but provides more accurate distance 
```

PR #659, opened 2025-05-19
Add DB migration manager, remove non-duplicate geometries requirement from transit DB, make transit walk links cost a tiny amount

Jake-Moss:
Bit of a mixed PR but this adds a DB migration manager to allow us to patch databases as we go, migrations are applied via the in order by their ID. Out of order migration application will raise an RuntimeError if detected. Adds a `migrations` table to track the ones that have been applied. Because we ship the migrations with the code you can't have a migration that is compatible with the code, except when downgrading. ATM downgrades are not considered at all, I might add some method to detect unknown migrations but at that point you're on your own anyway.

This should fix the zero cost transit links as well @pedrocamargo .

### TODO
- [x] Documentation, both comments and a proper page


tests/aequilibrae/project/test_migration_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    def setUp(self):
+        self.tmpdir = tempfile.TemporaryDirectory()
+        self.connection = sqlite3.connect(":memory:")
+
+        self.migrations_file = pathlib.Path(__file__).parent.parent.parent / "data" / "mock_migrations" / "init.py"
+        self.migrations_duplicate = (
+            pathlib.Path(__file__).parent.parent.parent / "data" / "mock_migrations" / "duplicate_init.py"
+        )
+        self.migrations_negative = (
+            pathlib.Path(__file__).parent.parent.parent / "data" / "mock_migrations" / "negative_init.py"
```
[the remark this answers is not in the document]
Jake-Moss: Not an issue

docs/source/useful_links/version_history.rst
```diff
@@ -214,6 +214,11 @@ In the meantime, you can find the documentation for all versions since 0.5.3.
 
 .. grid::
```
pedrocamargo, line 215: @r-akemii , is this formatted correctly?

pedrocamargo:
@Jake-Moss , this looks awesome, but I didn't have time to finish the review today, unfortunately.

One thing that occurs to me:  Is this the time to add support for saving graphs for more than one period to be kept in the DB?

docs/source/useful_links/version_history.rst
r-akemii: It's missing an admonition. Otherwise, it will appear as two grids in a row rather than three.

docs/source/useful_links/version_history.rst
```diff
@@ -214,6 +214,11 @@ In the meantime, you can find the documentation for all versions since 0.5.3.
...
 
+    .. grid-item-card:: 1.3.2
+        :link:  https://www.aequilibrae.com/docs/python/V.1.3.2/
+        :link-type: url
+        :text-align: center
+
```
r-akemii, line 221:
```suggestion

    .. grid-item::
     
```

docs/source/useful_links/version_history.rst
Jake-Moss: Whoops, thanks!

Jake-Moss:
> One thing that occurs to me: Is this the time to add support for saving graphs for more than one period to be kept in the DB?

We certainly could. I'm imagining it could be done in three ways:
 - Add another column to the links table `periods`, similar to `modes`, it's a string that is comma separated integers as we have been using `1` for the default period id. We could also switch to using characters, as we do with modes. This would allow sharing links between periods, but the transit graph builder isn't setup for this at all. So we should just duplicate it on insertion then have some deduplication procedure to collapse identical links modulo period id.
 One issue I can imagine with this deduplication is that the link_id would change on insertion, saving the graph would require that we reread it (if it's done as a trigger), or do the deduplication in Python.
- Add another `period` column, it specifies a single period id for the link. Essentially the same as the above but we just say that a link can only belong to single period.
 - Create a new links table for each period, the period id then dictates which table is used. We can't share links here but everything would be separate and isolated. I think this would be simpler to implement than the first but then we can't share links between periods, and since the links table is a geo table we would create more spatial tables. 
 Optional we could move the geometry off onto another table keyed by link_id and period id.

pedrocamargo (approved):

PR #672, opened 2025-05-28
Align PT origin/od node IDs with their respective zone ID

Jake-Moss:

PR #674, opened 2025-06-10
Fix run module import, fix PT skim column names, fix new field editor fiels not showing in QGIS

Jake-Moss:
Insert the loaded module into sys.modules so we can use relative imports. I think I removed this after seeing Jamie’s version but this break modules that use relative imports.

This also pulls over a commit from #672 which we need, to fix the RC column names.
This also fixes fields added via the field editor not appear in the QGIS by updating the geo column stats table.

### DIFF
--- a/aequilibrae/paths/route_choice.py
+++ b/aequilibrae/paths/route_choice.py
@@ -500,7 +500,8 @@ class RouteChoice:
             added_dfs.append(pd.DataFrame({(*k, "tot"): df[(*k, "ab")] + df[(*k, "ba")]}))
 
         df = pd.concat([df] + added_dfs, axis=1).set_index("link_id")
-        df.columns = pd.MultiIndex.from_tuples(df.columns)
+        df.columns = ["_".join(x) for x in df.columns]
+
         return df.sort_index()
 
     def set_select_links(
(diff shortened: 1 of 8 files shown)


aequilibrae/paths/route_choice.py
```diff
@@ -500,7 +500,8 @@ def __link_loads_to_df(self, mapping, lids, link_loads):
             added_dfs.append(pd.DataFrame({(*k, "tot"): df[(*k, "ab")] + df[(*k, "ba")]}))
 
         df = pd.concat([df] + added_dfs, axis=1).set_index("link_id")
-        df.columns = pd.MultiIndex.from_tuples(df.columns)
+        df.columns = ["_".join(x) for x in df.columns]
```
jamiecook, line 503: just wanted to check if we are sure this is a multi-element column name at this point? If they are strings this will still run, but in an unexpected way: 

```python
>>> "_".join("hello")
'h_e_l_l_o'
```

pedrocamargo: @Jake-Moss , I think this is the opportunity to change column names in the route choice results to be in the same standard as traffic assignment.

pedrocamargo: @jamiecook , what @Jake-Moss is doing here is solving the issue I was referring to, which is having all result column names being tuples in the route choice module

pedrocamargo (approved):

PR #676, opened 2025-06-19
Case robustness

pedrocamargo:
Upper case columns in the graph causes the detection of fields that are AB/BA to fail.


aequilibrae/paths/graph.py
```diff
@@ -634,6 +634,11 @@ def create_compressed_link_network_mapping(self):
 
         return create_compressed_link_network_mapping(self)
 
+    def __setattr__(self, key, value):
+        if key == "network" and isinstance(value, pd.DataFrame):
+            value.columns = [col.lower() for col in value.columns]
+        self.__dict__[key] = value
```
Jake-Moss: This is just personal taste but I don't like modifying `__dict__` directly because some objects don't have a `__dict__`, but in this case we definitely do
```suggestion
        super().__setattr__(key, value)
```

Jake-Moss (approved):

Jake-Moss:
Not sure what is causing these failures but I can reproduce them locally when on this branch...

PR #677, opened 2025-06-27
Improvements to progress bars

pedrocamargo:


pedrocamargo:
@Jake-Moss , I was looking into ways to have progress bars for Route Choice and PT assignment, and came across this alternative

What do you think about it?
[Cython prange progress bar.pdf](https://github.com/user-attachments/files/20956259/Cython.prange.progress.bar.pdf)

aequilibrae/paths/cython/route_choice_set.pyx
```diff
@@ -375,8 +379,12 @@ cdef class RouteChoiceSet:
                             b_nodes_matrix[thread_id],
                             self.b_nodes_view,
                         )
+                    with gil:
+                        with cython.parallel.atomic():
+                            progress[0] += 1
```
Jake-Moss: I don't think the `cython.parallel.atomic()` context manager exists. Cython itself doesn't have any atomic operations or threading helpers besides prange and parallel.

Jake-Moss:
What we can do is spin off another Python thread before we start the prange blocks which then periodically acquires the gil to update the progress bar. We can use an atomic counter from the C++ stdlib and have all threads increment that. This would be essentially what you have here

aequilibrae/utils/cython/atomic_signal.pyx
```diff
@@ -0,0 +1,77 @@
...
+
+    @property
+    def total(self):
+        return self.__total
+
+    @total.setter
+    def total(self, val: int):
+        self.__total = val
+        if self.__signal.pbar is not None:
+            self.__signal.pbar.total = val
```
Jake-Moss: @r-akemii Is there signal that we could send to QGIS to set/update the maximum for the progress bar?

For context, when using the function decorator version of this class, the total may not be known before the function executes so I allow the user to set the total after it has been constructed via the `total` property. Because the progress bar is created in the other thread, it may be created before the user has set the `total` attribute so it's possible for this to be 0 for the first display, but updated next iteration. How can we make this work in QGIS? ATM it assumes that SIGNAL always has a `pbar` attribute.

Jake-Moss:
@pedrocamargo I've added an SIGNAL wrapper written in Cython using C++ atomic operations and added it to RC. This is basically what you had. It spins off another thread that sleeps without the gil, then acquires it to update the TQMD progress bar. This should be pretty light weight but it does need the gil to do the print outs. I'm not sure if theres a way around that.

aequilibrae/utils/python_signal.py
```diff
@@ -14,7 +14,7 @@
     else:
         from tqdm import tqdm  # type: ignore
 
-show_status = os.environ.get("AEQ_SHOW_PROGRESS", "TRUE") == "TRUE"
+show_status = os.environ.get("AEQ_SHOW_PROGRESS", "TRUE") == "TRUE" or missing_tqdm
```
Jake-Moss: If we're missing tqdm then surely we can't show a progress bar? Or is this for QGIS?
```suggestion
show_status = os.environ.get("AEQ_SHOW_PROGRESS", "TRUE") == "TRUE" and not missing_tqdm
```

aequilibrae/utils/cython/atomic_signal.pyx
r-akemii: @Jake-Moss I think that if you emit two different "start" signals, it should work in QGIS. The maximum of the progress bar is set using the "start" signal, so you could do something like start the function without knowing the total progress, finish it, and start the progress bar once again, knowing the progress bar maximum. 

Try to keep the action format as the one in the class `PythonSignal`. This way, I think you won't have any problems with QGIS.

Jake-Moss: Thanks, it works with TQDM so I'm happy with that.

pedrocamargo:
Since we Both worked on this, @Jake-Moss , let's wait for @jamiecook to review it.

aequilibrae/utils/python_signal.py
```diff
@@ -63,8 +63,9 @@ def emit(self, val):
 
         elif action == "set_text":
             desc = str(val[1]).ljust(50)
-            if self.pbar is not None and self.pbar.desc != desc:
-                self.pbar.set_description(desc, refresh=True)
+            if self.pbar is not None:
+                if self.pbar.desc != desc:
+                    self.pbar.set_description(desc, refresh=True)
```
jamiecook:
```suggestion
            if self.pbar is not None and self.pbar.desc != desc:
                self.pbar.set_description(desc, refresh=True)
```

PR #678, opened 2025-07-09
Add Results and ResultRecord objects for easier result manipulation

Jake-Moss:
This adds a similar interface for the results as we have for the matrices. 

Results should be set in a manner similar to
```python
        report = {"setup": self.info()}
        record = project.results.new_record(
            table_name=table_name,
            procedure="transit assignment",
            procedure_id=self.procedure_id,
            procedure_report=json.dumps(report),
            timestamp=self.procedure_date,
            description=self.description,
        )
        record.set_data(df)
```
where `df` is the DataFrame to save as the table.

Results can be read with `project.results.get_results("table_name")`.

This PR also includes a migration to move the result record entries from the transit database to the project database. Previously we had two sets of records that may have conflicted with each other.


aequilibrae/project/data/result_record.py
```diff
@@ -0,0 +1,129 @@
...
+
+        self.__dict__["_project_conn"] = project_conn
+        self.__dict__["_results_conn"] = results_conn
+
+    def save(self) -> None:
+        """Saves results record to the project database.
+
+        Creates a new record if it doesn't exist or updates an existing one.
+        """
+        with self._project_conn or self.project.db_connection as conn:
```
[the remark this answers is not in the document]
Jake-Moss: The `or` is evaluated before the context manager protocol is used. Just when I thought this bot were getting good, it says stupid stuff like this.

https://docs.python.org/3/reference/compound_stmts.html#the-with-statement
> ```
> with_stmt          ::= "with" ( "(" with_stmt_contents ","? ")" | with_stmt_contents ) ":" suite
> with_stmt_contents ::= with_item ("," with_item)*
> with_item          ::= expression ["as" target]
> ```
> 1. The context expression (the expression given in the [with_item](https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-with_item)) is evaluated to obtain a context manager.

aequilibrae/project/data/result_record.py
```diff
@@ -0,0 +1,129 @@
...
+            sql = "SELECT COUNT(*) FROM results WHERE table_name=?"
+
+            if conn.execute(sql, [self.table_name]).fetchone()[0] == 0:
+                data = [
+                    str(self.table_name),
+                    str(self.procedure),
+                    str(self.procedure_id),
+                    json.dumps(self.procedure_report),
```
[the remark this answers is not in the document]
Jake-Moss: Reasonable catch

Jake-Moss:
Just thought of some minor issues when applying the migration to a new database that doesn't have a results table, and need to update the reference files.

aequilibrae/project/database_specification/transit/tables/results.sql
```diff
@@ -1,25 +0,0 @@
...
---@ which is created at runtime.
---@
---@ The **procedure_report** field holds the output of the complete procedure report. 
---@
---@ The **timestamp** field holds the information when the procedure was executed.
---@
---@ The **description** field holds the user-provided description of the result.
-
-
-create TABLE if not exists results (table_name       TEXT     NOT NULL PRIMARY KEY,
```
pedrocamargo, line 20: Weren't we going to add a couple more columns to the results table in the project database?

Jake-Moss: Done

aequilibrae/project/data/result_record.py
```diff
@@ -0,0 +1,118 @@
...
+                data = [
+                    str(self.table_name),
+                    str(self.procedure),
+                    str(self.procedure_id),
+                    self.procedure_report,
+                    str(self.timestamp),
+                    str(self.description),
+                ]
+                conn.execute(
+                    "INSERT INTO results (table_name, procedure, procedure_id, procedure_report, timestamp, description)"
```
pedrocamargo: Add scenario & year to SQL.
Make it insert or update

aequilibrae/project/data/result_record.py
```diff
@@ -0,0 +1,118 @@
...
+    def get_data(self) -> pd.DataFrame:
+        """Returns the results data for further computation.
+
+        Returns:
+            **df** (:obj:`pd.DataFrame`): DataFrame containing the results data.
+        """
+        with self._results_conn or self.project.results_connection as conn:
+            return pd.read_sql(f"SELECT * FROM {self.table_name}", conn)
+
+    def set_data(self, df: pd.DataFrame, **kwargs) -> None:
```
pedrocamargo, line 90: Save record here

aequilibrae/project/database_specification/network/tables/results.sql
```diff
@@ -22,4 +22,6 @@ create TABLE if not exists results (table_name       TEXT     NOT NULL PRIMARY K
                                     procedure_id     TEXT     NOT NULL,
                                     procedure_report TEXT     NOT NULL,
                                     timestamp        DATETIME DEFAULT current_timestamp,
-                                    description      TEXT);
+                                    description      TEXT,
+                                    year             TEXT,
```
pedrocamargo: **Scenario** & **Year** need to come as the first two columns in the database

pedrocamargo (approved):

PR #679, opened 2025-07-17
Support multiple scenarios

Jake-Moss:
This is the start of basic scenario support is aequilbrae, it functions by changing where all the file paths for the databases and matrices point to.

ATM scenarios may be created as blank projects, or clones of the active scenario. The "root" scenario is treated as a special case, it is referred to by the name "root" and exists in all projects (once upgraded).

Scenarios can be "switched to" with the `project.switch_scenario` function.

I think this is the least intrusive and backward-compatible way, other methods I've tried require some sort of additional "project management"


aequilibrae/transit/transit.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, project):
             **project** (:obj:`Project`, *Optional*): The Project to connect to. By default, uses the currently
             active project
         """
-        WorkerThread.__init__(self, None)
+        super().__init__(self, None)
```
r-akemii: Do we need to change the `WorkerThread` for `super`? I believe it might affect QAequilibraE.

aequilibrae/project/project.py
```diff
@@ -266,15 +308,110 @@ def __create_empty_network(self):
...
+            pass
+
+        try:
+            shutil.copyfile(
+                self.project_base_path / "results_database.sqlite", scenario_path / "results_database.sqlite"
+            )
+        except FileNotFoundError:
+            pass
+
+        shutil.copy(self.parameters.file, scenario_path)
```
r-akemii: Instead of using `try/except` couldn't we use `os.path.exists` to check if the file exists and if so, copy it to the new folder?

r-akemii:
@Jake-Moss, sorry if it's a silly question, but do we guarantee that when changing scenarios, we close the prior database connections safely? Is that any chance that a change in any of the current scenarios database indirectly affects the root databases?

aequilibrae/transit/transit.py
Jake-Moss: In this case they do the same thing, it's just better Python practice to use `super()` over manually specifying the class name. `super()` will return a proxy object for the next parent or sibling class, specifically it will follow the [method resolution order](https://docs.python.org/3/howto/mro.html#python-2-3-mro). In our case it's a simple straight line, but for more complicated things like multiple inheritance, it's hard to get that right.

aequilibrae/project/project.py
Jake-Moss: Certainly could, largely just a difference in style. The one you described is called "look before you leap", and the try-except is more "ask for forgiveness". One small difference is the try-except is helps mitigate [time of check to time of use ](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bugs, but that's not really a concern here

Jake-Moss:
> @Jake-Moss, sorry if it's a silly question, but do we guarantee that when changing scenarios, we close the prior database connections safely? Is that any chance that a change in any of the current scenarios database indirectly affects the root databases?

That's the main thing I'm concerned about. If some object holds onto a connection somehow then switching scenarios will break spectacularly. Everything should really use the connection properties on the `Project` object. I'm going to go and remove all uses of other database connections methods

pedrocamargo:
@Jake-Moss , I did a bit more investigation here and it DOES look like a segfault in one of the Cython extensions, but it simply stops running on Windows. No error message, no stack trace. I tried a few things and nothing changed that

aequilibrae/project/project.py
```diff
@@ -82,35 +92,35 @@ def open(self, project_path: str) -> None:
...
 
     @property
     def about(self) -> About:
-        return self.__about
+        return self.scenario.about
 
     @property
     def logger(self) -> logging.Logger:
-        return self.__logger
+        return self.scenario.logger
```
pedrocamargo, line 107: We need to think what do we want to do with the logger.  One per scenario?  One global plus one per scenario?

Jake-Moss: I'm not quite sure what is the best way to handle the logging, as a library all of our modules should get their own logger via `logging.getLogger(__name__)`,  then a user can configure the `"aequilibrae"` logger, as a sort of root logger for Aequilibrae. Each module having their own logger allows someone to also configure that particular module and all it's children. You could have the `"aequilibrae"` logger on the ERROR level, but the `"aequilibrae.traffic_assignment"` logger on the INFO level.

I think what we want is to use logging filters attached to logging handlers. We have a filter + handler setup for each scenario, then when executing there's some way to note which scenario the log is for, and only emit that log on the handler with the filter than matches that scenario.

https://docs.python.org/3/library/logging.html#filter

This seems like something that would be nice to do for 2.0.

aequilibrae/project/database_specification/network/tables/scenarios.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS scenarios (scenario_name TEXT UNIQUE NOT NULL, description TEXT);
+
+--#
+INSERT INTO 'scenarios' (scenario_name, description) VALUES('root', 'The default, and root, scenario for an AequilbraE project. The name "root" is treated as a special case.');
```
pedrocamargo, line 4: Should we add a trigger to prevent users from deleting the root scenario from this table

aequilibrae/project/project.py
```diff
@@ -82,35 +92,35 @@ def open(self, project_path: str) -> None:
 
     @property
     def project_base_path(self):
-        return self.__base_path
+        return self.scenario.base_path  # do we want to use the root scenario here?
```
pedrocamargo: In response to the comment, I am happy to keep the scenario path here. We do have the root_scenario always available.

Maybe we should rename _project_base_path_ to _base_path_ or maybe even keep _project_base_path_ as the actual root directory

aequilibrae/project/project.py
```diff
@@ -82,35 +92,35 @@ def open(self, project_path: str) -> None:
...
 
     @property
     def path_to_file(self):
-        return self._project_database_path
+        return self.scenario.path_to_file  # do we want to use the root scenario here?
```
pedrocamargo: Here the answer is no.

aequilibrae/project/project.py
pedrocamargo: Can you add this as an issue for AequilibraE 2.0, @Jake-Moss

aequilibrae/project/project.py
pedrocamargo: Can you add this as an issue for AequilibraE 2.0, @Jake-Moss

aequilibrae/project/project.py
```diff
@@ -331,3 +346,93 @@ def __setup_logger(self):
             logger.addHandler(get_log_handler(self.project_base_path / "aequilibrae.log"))
 
         return logger
+
+    def list_scenarios(self):
+        with self.db_connection as conn:
+            return pd.read_sql("SELECT * FROM scenarios", conn)
+
+    def switch_scenario(self, scenario_name: str):
```
pedrocamargo: is _switch_scenario_ the best name for it?  

Would _use_scenario_ feel more natural? Genuine question here.

docs/source/examples/aequilibrae_project/plot_scenarios.py
```diff
@@ -0,0 +1,167 @@
...
+
+print(f"Current scenario network has {len(project.network.links.data)} links")
+print(f"Current scenario network has {len(project.network.nodes.data)} nodes")
+
+# %%
+# Creating new scenarios
+# ----------------------
+# We can create empty scenarios or clone existing ones
+
+# Create an empty scenario for testing modifications
```
pedrocamargo:
```suggestion
# Create an empty scenario to manually populate with a future/different network
```

docs/source/examples/aequilibrae_project/plot_scenarios.py
```diff
@@ -0,0 +1,167 @@
...
+project.list_scenarios()
+
+# %%
+# Switching between scenarios
+# ---------------------------
+# Each scenario operates independently with its own data
+
+# Switch to the cloned scenario
+project.switch_scenario("alternative_parameters")
+print(f"This scenario has {len(project.network.links.data)} links")
```
pedrocamargo, line 82: Let's add a little modification to this network.  Half the capacity for a couple of links, for example

docs/source/examples/aequilibrae_project/plot_scenarios.py
```diff
@@ -0,0 +1,167 @@
...
+mat.computational_view()
+
+# Create traffic assignment with alternative parameters
+assigclass = TrafficClass("car", graph, mat)
+assignment = TrafficAssignment(project)
+assignment.add_class(assigclass)
+assignment.set_vdf("BPR")
+
+# Use different VDF parameters than we might use in the root scenario
+assignment.set_vdf_parameters({"alpha": 0.25, "beta": 6.0})  # More congested parameters
```
pedrocamargo: This is not something people do in practice, hence the suggestion above to change the network itself.

docs/source/examples/aequilibrae_project/plot_scenarios.py
```diff
@@ -0,0 +1,167 @@
...
+alt_results = project.results.list()
+print(f"Alternative scenario has {len(alt_results)} result tables")
+
+# Each scenario maintains its own results database
+alternative_assignment_exists = "alternative_assignment" in alt_results["table_name"].values
+print(f"Alternative assignment result exists in this scenario: {alternative_assignment_exists}")
+
+# %%
+# Best practices for scenario management
+# --------------------------------------
```
pedrocamargo, line 151: Excellent note

aequilibrae/project/database_specification/network/tables/scenarios.sql
Jake-Moss: Yep good idea

aequilibrae/project/project.py
Jake-Moss: Agreed, I've changed it to be `use_scenario`

docs/source/examples/aequilibrae_project/plot_scenarios.py
Jake-Moss: Yep fair, will make that change

docs/source/examples/aequilibrae_project/plot_scenarios.py
Jake-Moss: Done

pedrocamargo (approved):

pedrocamargo:
Well done, @Jake-Moss !

PR #681, opened 2025-07-22
Revamps tests

pedrocamargo:


pedrocamargo:
@Jake-Moss , I am being a little brave and forcing a merge here.

PR #684, opened 2025-07-29
Removes permanently opened OMX files

pedrocamargo:


Jake-Moss (approved):
Looks nice. I never realised that `tables.File` has a context manager and thus so `omx.File`. Will definitely be using that from now on

PR #691, opened 2025-08-05
Add NumPy 2.0+ compatibility support

onlyname0:
Update version requirements and fix compatibility issues to support NumPy 2.0 and later versions while maintaining backward compatibility.

Changes:
- Update version constraints in pyproject.toml, conda_forge_recipe.yaml, and Pipfile
- Fix uint8 arithmetic overflow in matrix operations by converting to Python int
- Replace deprecated np.object_ with object type
- Fix numpy scalar string representation in matrix totals
- Update deprecated array slice assignments in GTFS loader
- Fix chained assignment in CSV parser with np.where()
- Replace deprecated np.core.* with np.char and np.rec
- Remove unnecessary test return value

All 378 tests pass with both NumPy 1.x and 2.x versions. Deprecation warnings reduced from 117 to 42.


pedrocamargo:
@Jake-Moss , Let's move this PR to a working branch for AequilibraE 2.0?
Let's also restore the production of QGIS-specific binaries. It is much easier to handle things that way.

Pipfile
```diff
@@ -26,7 +26,7 @@ matplotlib = "*"
 pytest-xdist="*"
```
pedrocamargo, line 26: This file should be dropped entirely

pyproject.toml
```diff
@@ -77,7 +77,7 @@ tests = [
   "wheel",
 ]
 
-build=["setuptools", "numpy<1.99", "cython", "wheel"]
```
pedrocamargo, line 80: We need smarter changes here.  PyTables requires NumPy<2 for Python 3.9 and 3.10. For main dependencies

    "numpy<1.99; python_version <= '3.10'",
    "numpy>=2.0; python_version >= '3.11'",

Pipfile
Jake-Moss: We'll just drop this later, it's unused for now

pyproject.toml
Jake-Moss: Done

Jake-Moss (approved):
Thank you for this @onlyname0 , changes look good to me

PR #697, opened 2025-08-08
Enhancement: Saving list of disconnected centroids for network debugg…

BishoyKelleny:
This block improves error-checking and debugging for network preparation by verifying that all defined centroids are present in the node list of the graph. Specifically, if any centroids are missing, it:
Issues a warning via warnings.warn.
Identifies and lists the missing centroid IDs.
Saves them to a CSV file named missing_centroids_from_graph.csv in the current working directory.
Prints the file location to help the user locate and inspect the issue.
This helps ensure network integrity and provides a convenient mechanism for diagnosing missing centroids during model setup or troubleshooting.

### DIFF
--- a/aequilibrae/paths/graph.py
+++ b/aequilibrae/paths/graph.py
@@ -249,6 +249,13 @@ class GraphBase(ABC):  # noqa: B024
         present_centroids = np.isin(centroids, nodes, assume_unique=True)
         if not present_centroids.all():
             warnings.warn("Found centroids not present in the graph!\n" + str(centroids[~present_centroids]))
+                        missing_centroids = centroids[~np.isin(centroids, nodes)]
+            warnings.warn(f"Found {len(missing_centroids)} centroids not present in the graph!")
+            # Save to disk for network debugging and fixing
+            output_path = os.path.join(os.getcwd(), "missing_centroids_from_graph.csv")
+            pd.Series(missing_centroids).to_csv(output_path, index=False, header=False)
+            print(f"Missing centroids written to: {output_path}")
+            
         nodes = np.setdiff1d(nodes, centroids, assume_unique=True)
         all_nodes = np.hstack((centroids, nodes)).astype(self.__integer_type)
 


aequilibrae/paths/graph.py
```diff
@@ -249,6 +249,13 @@ def _build_directed_graph(self, network: pd.DataFrame, centroids: np.ndarray):
         present_centroids = np.isin(centroids, nodes, assume_unique=True)
         if not present_centroids.all():
             warnings.warn("Found centroids not present in the graph!\n" + str(centroids[~present_centroids]))
+                        missing_centroids = centroids[~np.isin(centroids, nodes)]
+            warnings.warn(f"Found {len(missing_centroids)} centroids not present in the graph!")
+            # Save to disk for network debugging and fixing
+            output_path = os.path.join(os.getcwd(), "missing_centroids_from_graph.csv")
+            pd.Series(missing_centroids).to_csv(output_path, index=False, header=False)
+            print(f"Missing centroids written to: {output_path}")
+
```
jamiecook, line 258: This looks beneficial, but there is already a warning logged which this duplicates. 

```suggestion

            missing_centroids = centroids[~np.isin(centroids, nodes)]
            num_missing = len(missing_centroids)
            
            # Save to disk for network debugging and fixing
            output_path = os.path.join(os.getcwd(), "missing_centroids_from_graph.csv")
            pd.Series(missing_centroids).to_csv(output_path, index=False, header=False)
            warnings.warn(f"Found {num_misisng} centroids not present in the graph!\nFull list available in {output_path}\n") + str(centroids[~present_centroids].head()))
```

pedrocamargo:
Thanks for the contribution, @BishoyKelleny .  My biggest concern with this is the saving of the missing centroids:
* Simply saving to the current folder is not a good practice, as this may not work for everyone's workflow
* In some systems, this may not work at all, as the user would not have access/permission to save in the current folder.

A reasonable alternative would be to have an additional argument (that defaults to FALSE) to allow users to return the missing centroids if they are missing, or even a separate function for that purpose.

pedrocamargo:
@BishoyKelleny , Any views on our comments? If none, we can close this PR

pedrocamargo:
Closing, as no follow-up was provided

PR #700, opened 2025-08-11
Removes black in favor of a Ruff as the only linter

pedrocamargo:


pyproject.toml
```diff
@@ -156,13 +146,25 @@ exclude = [
...
-target-version = "py39"
+target-version = "py312"
+indent-width = 4
 
 [tool.ruff.lint.mccabe]
 max-complexity = 20
 
+[tool.ruff.format]
+
+quote-style = "double"
```
jamiecook, line 159: do we need to add the line-length limit here to match what was being done in black....

pyproject.toml
```diff
@@ -156,13 +146,25 @@ exclude = [
     ".venv311",
     ".venv312",
 ]
-line-length = 120
```
jamiecook, line 159: actually ... it's removed from here... going back to 88 char width?

jamiecook (approved):

pyproject.toml
pedrocamargo: I missed that!!!

pyproject.toml
pedrocamargo: Done

Jake-Moss (approved):

PR #720, opened 2025-09-05
Missing references on API docs

r-akemii:
In this PR, we fix the missing references in API docs and correct docstring errors when rendering the docs.


Jake-Moss:
I feel like the API reference should try to include everything rather than just a selection that we have to keep up-to-date. I've done a completely auto generated one before so I'm going to try and apply it here as well. We can just revert the commits late if we're not happy with it

Jake-Moss:
Went down a too-long rabbit hole with my custom templates that I'd used previously. Seems like just adding the `:recursive:` tag and linking to a different page gets 95% of the way there and is much simpler.

Jake-Moss (approved):

Jake-Moss:
Given I've also now changed how the docs a generate, I think a quick review from @ pedrocamargo would also be good.

r-akemii:
@Jake-Moss the `:recursive:` tag you added is really nice! I personally don't like how it looks when rendering (it's just a matter of embellishment), but it does save a lot of work when adding new modules and is a significant improvement. Before merging the PR, I'm gonna check if the API references in the examples are redirecting to the correct page.

Jake-Moss:
> I personally don't like how it looks when rendering (it's just a matter of embellishment)

Yeah it is a little bit of a downgrade visually, on the top module page at least. It's a little bare. Unfortunately adding prose to the auto generated pages from separate .rst files doesn't seem possible without custom templates, and even then it's a bit of a pain. I think if we added module docs strings they would be rendered on those pages.

PR #726, opened 2025-10-13
AequilibraE CLI

r-akemii:
In this PR we add the AequilibraE CLI and its configuration to run all entry points in the run module.

### DIFF
--- a/aequilibrae/utils/aeq_cli.py
+++ b/aequilibrae/utils/aeq_cli.py
@@ -12,7 +12,7 @@ from aequilibrae.project import Project
 @click.pass_context
 def cli(ctx):
     if ctx.invoked_subcommand is None:
-        click.echo("You can only run procedures from the 'run' module.")
+        click.echo("You can only invoke commands run or list-functions")
 
 
 def available_parameters():
@@ -42,6 +42,14 @@ def run(project_dir, function, params):
     click.echo(result)
 
     project.close()
+
+
+@cli.command(help="List available functions in the run module.")
+@click.option("-d", "--project_dir", required=True, help="Project directory", default=os.getcwd)
+def list_functions(project_dir):
+    p = Parameters(Path(project_dir))
+
+    click.echo("Available functions: ", list(p.parameters["run"].keys()))
 
 
 if __name__ == "__main__":
--- a/aequilibrae/utils/aeq_cli.py
+++ b/aequilibrae/utils/aeq_cli.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+import os
+import enum
+import yaml
+
+import click
+
+from aequilibrae.project import Project
+from aequilibrae.parameters import Parameters
+
+
+@click.group(invoke_without_command=False)
+@click.pass_context
+def cli(ctx):
+    if ctx.invoked_subcommand is None:
+        click.echo("You can only run procedures from the 'run' module.")
+
+
+def available_parameters():
+    current_dir = os.getcwd()
+
+    p = Parameters(Path(current_dir))
+    return p.parameters["run"]
+
+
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -88,6 +88,9 @@ build = [
 
 dev = ['aequilibrae[build, tests, docs]']
 
+[project.scripts]
+aeq = "aequilibrae.utils.aeq_cli:cli"
+
```
Jake-Moss, line 93: Missing direct `click` dependency. Currently black requires it but we don't.

```
~/O/aequilbrae-nix/aequilibrae λ uv pip show click
Using Python 3.12.7 environment at: /home/jake/OuterLoop/aequilbrae-nix/.venv
Name: click
Version: 8.2.1
Location: /home/jake/OuterLoop/aequilbrae-nix/.venv/lib/python3.12/site-packages
Requires:
Required-by: black
```

Jake-Moss:
I really like this, but is click required? Our usage seems to be pretty simple. If we could use the stdlib [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) instead I think that would be better.

Currently this only supports calling a single function with string only arguments, do we want to provide some mechanism for specifying how to parse other arguments for functions? 
With argparse, the way to do this is by providing a [sub-parser](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers), which is another argparse object but under a subcommand. We could include a simple sub-parser in the run module as a default under a special name. For example we could have the `aeq run` command be controlled by a parser defined in the run-module. It would allow shipping custom cli interfaces (with help and custom parsing) with a project.
I'm not sure how to do this with click by I'm sure it's possible.

For a non-customised function argument parsers, Claude spat out this

```python
import argparse
import inspect

def create_parser_from_function(func):
    """Create an argparse parser from a function's signature."""
    parser = argparse.ArgumentParser(description=func.__doc__)
    
    # Get function signature
    sig = inspect.signature(func)
    
    for param_name, param in sig.parameters.items():
        # Determine if parameter has a default value
        if param.default == inspect.Parameter.empty:
            # Required positional argument
            parser.add_argument(param_name, type=type(param.annotation) if param.annotation != inspect.Parameter.empty else str)
        else:
            # Optional argument with default
            parser.add_argument(
                f'--{param_name}',
                default=param.default,
                type=type(param.default) if param.default is not None else str
            )
    
    return parser

# Example function
def process_data(input_file: str, output_file: str, verbose: bool = False, max_items: int = 100):
    """Process data from input file to output file."""
    print(f"Processing {input_file} -> {output_file}")
    print(f"Verbose: {verbose}, Max items: {max_items}")

# Create and use parser
parser = create_parser_from_function(process_data)
args = parser.parse_args()

# Call function with parsed arguments
process_data(**vars(args))
```

which works well enough. We could instead inspect the `project.run` functions pulling defaults from the partial functions then looking at the signatures of the functions. This could create some default parser with function specific sub-parsers in the run module.

aequilibrae/utils/aeq_cli.py
```diff
@@ -0,0 +1,58 @@
...
+def cli(ctx):
+    if ctx.invoked_subcommand is None:
+        click.echo("You can only invoke commands run or list-functions")
+
+
+def available_parameters():
+    current_dir = os.getcwd()
+
+    p = Parameters(Path(current_dir))
+    return p.parameters["run"]
```
Jake-Moss, line 23: This uses the cwd on run, rather than using the directory provided. When run outside a project it prints
```
No pre-existing parameter file exists for this project. Will use default
No pre-existing parameter file exists for this project. Will use default
``` 
which isn't a useful output and might be misleading.

Jake-Moss:
Superseded by #729

PR #730, opened 2025-10-23
Bridge Cython/C++ and Python logging, support Cython/C++ interruptions, rework Cython/C++ progress bars

Jake-Moss:
This adds support for: 
 - logging from Cython and C++ in a unified manner to the Python loggers, and
 - interrupting long running Cython and C++ processes via the `cysignals` library.

It also refactors how progress bar support was implemented in #677 and overhauls how aequilibrae uses loggers. This PR supersedes #677. 

The general overview is that this introduces a `Bridge` object that spins off another thread which enters a sleep-wake loop and periodically collects logs from Cython/C++ and submits them to Python, checks pending signals and asks threads to exit gracefully, and updates progress bars.

Goals for this were:
 - Logging:
   - [X] Unified methods to submit logs from Cython and C++.
   - [X] Logs should be funnelled through the Python logging library.
   - [X] Delayed string formatting, we shouldn't format a log string unless we are going to log it.

 - Interrupt long running Cython/C++ processes
   - [X] We should respond to signals such as `SIGINT` in an appropriate manner, i.e. we should attempt to exit and return control to the Python process.
   - [X] Information about the signal should be bubbled up to Python, if `SIGINT` was sent a `KeyboardInterrupt` exception should be raised.

 - Progress bars:
   - [X] We should have the ability to report progress from Cython and C++ if appropriate.
   - [ ] Progress bars should function in both QGIS and terminal.

As a benefit of depending on `cysignals` we now get trace-backs when crashes in Cython/C++ occur. 

### Todo:
 - [ ] Documentation:
   - [ ] developer usage guides are required, along with common pitfalls and notes on misleading error messages from the Cython compiler, and 
   - [x] notes on configuring the Python loggers and how they interact with the TQDM log handler redirection.
 - [ ] QGIS integration, particularly around how progress bars are created and how logs are displayed.
 - [ ] Tests:
   - [ ] it's a little tough to test signal related things but `cysignals` has some good tests I can look at,
   - [ ] not sure how I'd test progress bars,
   - [ ] logging tests.
 - [x] Port existing uses of `AtomicSignal` to `Bridge` and `Bar`.
 - [x] Get long running processes to exit when asked to.


r-akemii:
Hey @Jake-Moss a few thoughts regarding the QGIS integration:

1. About the QGIS logging - don't worry about it. A coulpe weeks ago I standardized the logging signals, and they're related to QGIS actions, not AequilibraE ones. If AequilibraE logs are working in the Project as they should, no need to worry about QGIS logging.
2. QGIS progress bars - I'm not sure, but do we have nested progress bars in AequilibraE? When using cython, the signals are going to work as they do today, with the same actions and so on (if you're planning to change this, make sure to keep using actions that PyQt can understand. Probably there will be no change in Qt5 to Qt6, but maybe it's worth checking)? As we discussed before, I think that adding a progress bar function to qaequilibrae.py should do, but I'm not sure about handling nested progress bars. We used to have nested bars that were class-specific, with each bar represented by an object in the UI. One other thing that could work, is also creating a progress bar class in QAequilibraE and calling it on the go (eg.: you hit run and the parent window freezes or has a "loading" indicator, and a child window with progress bars shows up. When the process finishes, both parent and child window closes).

Jake-Moss:
Thanks @r-akemii !

> About the QGIS logging - don't worry about it. A coulpe weeks ago I standardized the logging signals, and they're related to QGIS actions, not AequilibraE ones. If AequilibraE logs are working in the Project as they should, no need to worry about QGIS logging.

I've pushed some code yesterday that overhauled how the logging in AequilibraE functioned. It's still standard Python logging, but we use module specific loggers now and no longer pass loggers around, nor is there a "project" logger any more. I've still got some work do with the scenario specific logging which will handle the log files. This was required to keep the progress bars intact while logging to the terminal. I'll write some sort of developer note which details what has changed. I suspect that QGIS will have some sort of StringIO stream the logs get written to, if so it will be very easy to get the new loggers working.

> A coulpe weeks ago I standardized the logging signals

I just had a look and couldn't see commits related to this. Could you please link them? I'll make sure I haven't broken anything

> do we have nested progress bars in AequilibraE?

Not at the moment. I'm more than happy to hold off on multiple progress bar support. 

> When using cython, the signals are going to work as they do today, with the same actions and so on 

I've got no plans to change the signal object but I have changed when they are created. My understanding is that qaequilibrae needs to be able to find the signal objects some how. I'll take a look and propose some solutions because now I am just creating them when required with no way for qaequilibrae to access them. Perhaps we should just create them all at once before hand?

Jake-Moss:
Rebased on develop, but there's enough changes to the logging I think this will have to target 2.0

r-akemii:
Hey @Jake-Moss!

> I just had a look and couldn't see commits related to this. Could you please link them?

Basically they're on this previous pull request (https://github.com/AequilibraE/qaequilibrae/pull/498), but it all started at 479.

> I've got no plans to change the signal object but I have changed when they are created. My understanding is that qaequilibrae needs to be able to find the signal objects some how. I'll take a look and propose some solutions because now I am just creating them when required with no way for qaequilibrae to access them. Perhaps we should just create them all at once before hand?

Yeah, indeed. qaequilibrae needs to find the signals somehow. I think that the way signals are currently created in AequilibraE work fine for qaequilibrae, so maybe using a variable `signal = SIGNAL(object)` to indicate if the classes emit signals could be interesting.

I started updating progress bars in qaequilibrae to use a custom class. I think that maybe this will make easier for you to test these changes in QGIS.

Jake-Moss:
I'm going to merge this PR to the 2.0 branch with the note that the QGIS signal sharing issue is still outstanding. It's not forgotten but there are other PRs built atop this one.

PR #731, opened 2025-11-14
Set start method, add env var for spatialite load path

Jake-Moss:
I'm not quite sure why, but changing the process start method fixes the segfaults on MacOS https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

"fork" is the *only* option that works, both "spawn" (default on Windows and MacOS) and "forkserver" (default on all platform from 3.14) do not work.

Interesting "spawn" was made the default on MacOS because "fork" was causing crashes. https://bugs.python.org/issue?@action=redirect&bpo=33725

I've also added a environment variable to allow changing which spatialite is loaded. This can accept a full path to avoid setting variables like `DYLD_LIBRARY_PATH`. This is what I used on AWS `export AEQ_SPATIALITE_PATH=/opt/homebrew/lib/mod_spatialite.dylib`

### DIFF
--- a/aequilibrae/utils/spatialite_utils.py
+++ b/aequilibrae/utils/spatialite_utils.py
@@ -51,7 +51,7 @@ def _connect_spatialite(path_to_file: os.PathLike, missing_ok: bool = False):
 
 def load_spatialite_extension(conn: Connection):
     conn.enable_load_extension(True)
-    conn.load_extension("mod_spatialite")
+    conn.load_extension(os.environ.get("AEQ_SPATIALITE_PATH", "mod_spatialite"))
 
 
 def is_spatialite(conn):
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,7 @@
+from multiprocessing import set_start_method
+
+set_start_method("fork")
+
 import os
 import uuid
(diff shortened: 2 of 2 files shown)


Jake-Moss:
@tomstephen please give these changes ago and let me know if it works.

Jake-Moss:
Before merging we should abide by the docs of `set_start_method` https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method and only change the default on the required platforms, so we should also test `spawn` and `forkserver` on Linux and Windows.

I think the whole `spatialite_utils.py` file could do with a clean up as well, it's a little hard to follow, I think the "auto download dir" and "where to find spatialite" should be consolidated into "load spatialite from this path" and a "if spatialite doesn't exist (file not found exception)" and "auto download env var" is set, download and place it there. If nothing is set, let the OS attempt to find it

tomstephen:
> @tomstephen please give these changes ago and let me know if it works.

All tests now pass on my machine! Apple M5 running Tahoe 26.1

Jake-Moss:
There's been lots of changes since the macos testing branch was created, I think it's worth updating that one first (merge develop)

PR #732, opened 2025-11-17
Fix segfault on ARM MacOS caused by multiprocessing

tomstephen:
This PR fixes the segfault occurring on Apple Silicon when running tests caused by the default start method used by multiprocessing. As found by @Jake-Moss, we use `set_start_method("fork")` to address. Confirmed by running tests locally (Apple M5 chip).

Closes #731. Address spatialite env variable and macos testing changes separately.


aequilibrae/__init__.py
```diff
@@ -21,5 +21,16 @@
...
 # When updating the version, one must also update the docs/source/useful_links/version_history.rst file
 version = "1.5.0"
+
+# On macos, we start multiprocessing with 'fork' to avoid segfaults. Other platform defaults are fine
+if sys.platform == "darwin":
+    try:
+        set_start_method("fork")
+    except RuntimeError:
+        # start method has already been set
+        pass
```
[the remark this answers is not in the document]
Jake-Moss: The hallucination machine is hallucinating again. I can't see anything about fork without exec being deprecated, nor is that issue related to this at all.

I do agree that we should be louder about this failure to set the spawn method. We shouldn't prohibit running aequilibrae if we didn't set it (so except is good), but we should make it known that it may crash. I'll propose another change

Jake-Moss (approved):
Looks good! Thanks, I'll wait for CI before merging, it's not happy about something

Jake-Moss:
Might need some additional tweaks to get openmp installed and running https://cibuildwheel.pypa.io/en/stable/faq/#macos-library-dependencies-do-not-satisfy-target-macos

I think we should first try just brew install llvm and openmp and going from there. If that doesn't work we may need to compile it ourselves in the before install step.

I'd also be fine not building wheels for this until we also get testing sorted, so could just comment out the macos-latest target for now.

docs/source/useful_links/installation.rst
```diff
@@ -29,6 +29,15 @@ Installation
 
   The Windows App Store ships a version of Python that contains an sqlite dll that does not support the loading of extensions. This means that Spatialite will not be loaded, and therefore AequilibraE will not work properly.
 
+macOS
+^^^^^
+
+AequilibraE no longer provides wheel files for macOS. To build AequilibraE from source, you will need to:
```
Jake-Moss:
```suggestion
AequilibraE does not provide pre-built wheel files for macOS. When installing from PyPi, the source distribution will be used and the library will be compiled locally. AequilibraE can also be built from source. For both methods you will need to:
```

docs/source/useful_links/installation.rst
```diff
@@ -29,6 +29,15 @@ Installation
...
+
+1. Install LLVM: ``brew install llvm``
```
Jake-Moss:
```suggestion
1. Install LLVM or another C/C++ compiler with OpenMP support: ``brew install llvm``
```

Jake-Moss (approved):
LGTM, thanks @tomstephen

PR #735, opened 2025-11-24
Prevent "No project parameters" warning on import

### LINKED ISSUE #734: Merely importing Aequilibrae emits a warning (about default parameters)
**Purpose of use**
In Movici (https://github.com/nginfra/movici-simulation-core/) we use aequilibrae as a backend for traffic simulations. Users do not use aequilibrae directly but we expose it through an abstraction layer

**Describe the bug**
Merely importing aequilibrae triggers a warning `"No pre-existing parameter file exists for this project. Will use default"`. This is due to a `Parameters` being instantiated in `project/network/osm/os_params.py` on import. At that point there has not been an opportunity to set an active project, so this warning is really unhelpful, especially if a user does not import aequilibrae specifically, but uses an abstraction layer (ie. our case). Currently, the only way I've found to suppress this message is to set a filter to the `"aequilibrae"` logger to explicitly filter out this message. This is a very ugly solution. It's not possible to change the log level, since aequilibrae will override the level on import as the `global_logger`. 

Directly reading the default `Parameters` in `osm_param` also leads to the situation where any `"osm"` parameters set in a projects parameters files, are ignored in this module. In this case this means that a user seems unable to override the "accept_language" setting

**To Reproduce**
1. import aequilibrae
2. see warning being emitted 

**Expected behavior**
No warning is emitted on import of aequilibrae

 - OS: ubuntu
 - Python version 3.13
 - Version 1.5.0

**Additional context**

A solution would be to defer reading the osm parameters until they are being used. For example through a function that takes in a Parameters (optionally) and outputs a dictionary or dataclass containing the parameters.

this function could also be a method or `@property` of the Parameters class

If you want I could whip up a PR for this

Jake-Moss:
Closes #734

### DIFF
--- a/aequilibrae/parameters.py
+++ b/aequilibrae/parameters.py
@@ -78,7 +78,7 @@ class Parameters:
 
     def restore_default(self):
         """Restores parameters to generic default"""
-        self.parameters = self.default
+        self.parameters = deepcopy(self._default)
         self.write_back()
 
     @classmethod
--- a/aequilibrae/parameters.py
+++ b/aequilibrae/parameters.py
@@ -79,8 +79,7 @@ class Parameters:
         self.parameters = self.default
         self.write_back()
 
-    @classmethod
     @property
-    def default(cls):
-        with open(cls.file_default, "r") as yml:
+    def default(self):
+        with open(self.file_default, "r") as yml:
             return yaml.safe_load(yml)
--- a/aequilibrae/parameters.py
+++ b/aequilibrae/parameters.py
@@ -1,9 +1,9 @@
-import os
-import yaml
-from copy import deepcopy
 import logging
-import pathlib
+from pathlib import Path
 from typing import Optional
+
+import yaml
+
 from aequilibrae.context import get_active_project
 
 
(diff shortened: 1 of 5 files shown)


pedrocamargo:
@Jake-Moss , how about the text of the warning?  Wouldn't this change cause that part of the code to never be triggered?  In that case, shouldn't it be removed?

Jake-Moss:
> @Jake-Moss , how about the text of the warning? Wouldn't this change cause that part of the code to never be triggered? In that case, shouldn't it be removed?

It can still be hit if you construct a `Parameters` object without an active project, I can't think of a use case for this but I think it's reasonable to warn about using the defaults then.
```python
jake@work-laptop:~/OuterLoop/aequilbrae-nix/aequilibrae$ python
Python 3.12.7 (main, Oct  1 2024, 02:05:46) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from aequilibrae.parameters import Parameters
>>> p = Parameters()
No pre-existing parameter file exists for this project. Will use default
>>>
```

Alternatively we could require the path argument and make an alternate constructor for the default.

pedrocamargo (approved):

PR #736, opened 2025-12-02
Enable macos testing

tomstephen:
Now that macOS support has been fixed, we enable macOS testing.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -114,4 +114,4 @@ jobs:
         source .venv/bin/activate
         pytest tests/ --durations=50 --dist=loadscope -n 4 --random-order --verbose
       env:
-        DYLD_LIBRARY_PATH: /opt/homebrew/opt/:$DYLD_LIBRARY_PATH
+        DYLD_LIBRARY_PATH: /opt/homebrew/lib:$DYLD_LIBRARY_PATH
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -83,7 +83,7 @@ jobs:
     - name: Download spatialite on Windows
       if: matrix.os == 'windows-latest'
       run: |
-        source .venv/Scripts/activate
+        .venv/Scripts/activate
         python tests/setup_windows_spatialite.py
 
     - name: Runs test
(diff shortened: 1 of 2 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -39,33 +39,52 @@ jobs:
...
+        CC: /opt/homebrew/opt/llvm/bin/clang
 
     - name: Download spatialite on Linux
       if: matrix.os == 'ubuntu-latest'
       run: |
         sudo apt update
         sudo apt install libsqlite3-mod-spatialite
         sudo apt install -y libspatialite-dev
-        pip3 install spatialite
+        uv pip install spatialite
```
pedrocamargo: This can be removed
```suggestion
```

.github/workflows/unit_tests.yml
```diff
@@ -39,33 +39,79 @@ jobs:
...
       max-parallel: 20
     steps:
     - uses: actions/checkout@v4
     - name: Set Python environment
-      uses: actions/setup-python@v5
+      uses: astral-sh/setup-uv@v7
       with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+          python-version: ${{ matrix.python-version }}
```
Jake-Moss: This action doesn't install Python btw, this just installs UV. You can use `uv python install` 

The `python-version` sets the `UV_PYTHON` environment variable. https://github.com/astral-sh/setup-uv?tab=readme-ov-file#do-i-still-need-actionssetup-python-alongside-setup-uv

The runners come pre-installed with a bunch of different Python versions and I think some of them don't have sqlite extensions enabled

tomstephen: Understood, perfect thank you!

.github/workflows/unit_tests.yml
Jake-Moss: Unrelated to this PR but it's a shame that name is taken. We could package libspatialite as a python package to distribute it as a wheel instead of relying on the system installation. This would resolve the whole "mod_spatialite" issue, we just import some package and use a path it specifies.

Could use GitHub CI to compile and package it automatically across all platforms we need + others

Jake-Moss:
🎉 Finally, thank you for your help with this @tomstephen

Jake-Moss (approved):

PR #737, opened 2025-12-05
Drop Python 3.9 support, drop black, bump ruff, set tables minimum version

Jake-Moss:


Jake-Moss:
This should help with the segfault we are seeing on #736 , for some reason. This also simplifies the pytables-numpy dependency situation.

pedrocamargo (approved):

PR #740, opened 2025-12-12
Un-pin tables

Jake-Moss:


pedrocamargo (approved):

pedrocamargo (approved):

Jake-Moss:
I'm now not sure this works? This is an annoying mess. I was planning on un-pinning tables because anaconda doesn't supply tables > 3.9.1, but trying this locally, it's now broken?

PR #741, opened 2025-12-12
Drop compile time numpy dependency, unpin tables

Jake-Moss:
This favours stdint type declarations for fixed-width integers over numpy. float64/float32 still need to be double/float.

By not having any compile time dependency we can (hopefully) be rid of those 
```
numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject 
```
errors and the like. This also improves our comparability with different tables versions.

We don't strictly require the `numpy >=1.26` dependency but I don't know a true lower bound, and this seems reasonable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "numpy>=2.0", "cython", "wheel"]
+requires = ["setuptools", "cython", "wheel"]
 
 [project]
 authors = [{ name = "Pedro Camargo", email = "pedro@outerloop.io" }]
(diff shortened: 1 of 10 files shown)


Jake-Moss:
Seems to be working now

pyproject.toml
```diff
@@ -27,7 +27,7 @@ readme = "README.md"
 requires-python = ">=3.10" # match classifiers above and ruff version below
 
 dependencies = [
-  "numpy>=1.26",  # Required because of our "NPY_TARGET_VERSION"
+  "numpy>=1.26",
```
pedrocamargo: Do we even need to require a minimum NumPy version anymore?

pedrocamargo (approved):

pyproject.toml
Jake-Moss: Nope, but I don't know what our actual version lower bound is, it's probably `1.21.2` which was the first version to add 3.10 support. We can certainly remove this.

```suggestion
  "numpy",
```

Jake-Moss: We'll see how this goes, worst case we tell users to upgrade, then add a constraint back

PR #742, opened 2026-01-07
Adds configuration file generation support for Simwrapper

PuttiKoselig:
### PLAN
__________

## TASK

Automatically generate SimWrapper-ready visualization outputs
from an AequilibraE Project.

## NTD:
- inspect project contents (network + results tables)
- export helper CSV and VegaLite files
- generate a SimWrapper `dashboard.yaml`

### Core
- [x] `_create_directories()`
- [x] `_add_to_generated_files()`
- [x] `_dashboard_skeleton()`
- [x] `_build_dashboard_config()`
- [x] `write_yamls()`

### Panel/Layout builders
- [x] `_intro_row()`
- [x] `_stats_rows()`
- [x] `_entire_network_row()`
- [x] `_links_info_row()`
- [x] `_capacity_map_row()`
- [x] `_scenario_metric_map()`
- [x] `_metric_comp_row()`
- [x] `_delay_factor_comp_row()`
- [x] `_voc_comp_row()`

### Assignment convergence
- [x] `_parse_convergence_json()`
- [x] `_export_convergence_csv()`
- [x] `_write_convergence_vega_spec()`
- [x] `_assignment_convergence_plot()`

### Panel Classes
- [x] `SimwrapperPanel`
- [x] `ConvergencePanel`
- [x] `TilePanel`
- [x] `TextPanel`
- [x] `AequilibraEMapPanel`

## LATER
- [x] check if works for other projects
- [x] color consideration


Jake-Moss:
As mentioned, this might be nice to have accessible as a cli program. This can be setup via https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#console-scripts

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,123 @@
...
+        self.generated_files = {} 
+        self._create_directories()
+
+    def _create_directories(self):
+        """
+        Create output directory structure for simwrapper
+
+        Structure:
+        PROJECT-DIRECTORY/
+            data/               # Data files referenced by configs
```
jamiecook: I was hoping that everything could be pulled into a simwrapper sub-dir - there is lots of "data" in a standard modelling project - i'd like to have this not pollute the top level more than necessary

PuttiKoselig: this has been resolved in latest commits

aequilibrae/utils/generate_simwrapper_config.py
PuttiKoselig, line 1: This also fixes directory printed in yaml to include the output_dir, it will not work without this.

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -343,6 +343,110 @@ def _capacity_map_row(self):
...
+        """makes scenario comparison map for a network's performance metric'
+
+        :Arguments:
+            **title** (:obj:`str`): panel title
+            **results_table** (:obj:`str`): results table to join to links
+            **metric_column** (:obj:`str`): metric to use for link colouring
+            **legend** (:obj:`list`): legend def for the map
+            **data_range** (:obj:`list`): value range used for colour scale
+            **palette** (:obj:`str`, *Optional*): colour palette to use
+            **width_by_link_type** (:obj:`bool`, *Optional*): vary line width by link type????????? weird
```
tomstephen: I think the options we'd want for width are by link type, capacity or the metric we are showing. By link type, think of it showing highways etc as "bigger" roads than quiet surface streets. Capacity is analogous to this. Generic metric option *could* be helpful, whether that be a link attribute (number of lanes, etc) or a different result (like show delay factor with width AND colour)

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -343,6 +343,110 @@ def _capacity_map_row(self):
...
+        """
+
+        panel = AequilibraEMapPanel(
+            title=title,
+            height=10,
+            width=3,
+            center=self.center,
+            zoom=self.zoom,
+            projection="EPSG:32719",
```
tomstephen: This is UTM 19S, so suitable for something like Coquimbo. I wonder if it's actually being used by SimWrapper? I will investigate

aequilibrae/utils/simwrapper_panel.py
```diff
@@ -0,0 +1,119 @@
...
+        self.center = center
+        self.zoom = zoom
+        self.projection = projection
+
+        self.defaults = None
+        self.extra_databases = None
+        self.layers = {}
+        self.legend = None
+
+    def set_defaults(self, defaults_dict):
```
tomstephen: Could we have a default set of defaults? If we are controlling all styling manually, and it should not fall back to the default, so the default-default could be neon green or something obvious.

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,503 @@
...
+        # links layer styling
+        style = {
+            "lineColor": {
+                "column": metric_column,
+                "palette": palette,
+                "dataRange": data_range,
+            }
+        }
+
+        # link type by line width?? made optional bc weird, but example yaml does this
```
tomstephen: As above comment

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,503 @@
...
+
+        config = self._dashboard_skeleton() # based config
+
+        # dashboard rows
+        rows = {
+            "introRow": self._intro_row(),
+            "statsRow": self._stats_rows(),
+            "entireNetworkRow": self._entire_network_row(),
+            "linkTypeAndCapasityRow": self._links_info_row() + self._capacity_map_row(),
+            "delayFactorCompRow": self._delay_factor_comp_row(),
```
tomstephen: A project could have no results (no tables in results_database.sqlite, or results_database.sqlite doesn't exist), or it could have only one table of results, or it could have multiple. We want to handle each of these cases differently.

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,503 @@
...
+    def _has_nodes(self):
+        """Checks if project has a network with nodes"""
+        return True
+
+    def _has_zones(self):
+        """Checks if project has a network with nodes"""
+        return True
+
+    def _has_skims(self):
+        """Checks if project has a network with skims"""
```
tomstephen: See [here](https://www.aequilibrae.com/latest/python/useful_links/_autosummary/aequilibrae.project.data.matrices.html#aequilibrae.project.data.matrices.Matrices). Also, I think we should rename skims -> matrices as these can include matrices-that-arent-skims, such as the demand matrix. Let's discuss soon what visualization we can do here.

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,503 @@
...
+                "description": "insert description"
+            },
+
+            "layout": {}
+        }
+
+        return config
+
+    def _intro_row(self):
+        """ resturns project details text panel"""
```
tomstephen: Let's consider [this](https://www.aequilibrae.com/latest/python/useful_links/_autosummary/aequilibrae.project.about.html) info, if that draws a blank let's use the name of the folder (replace underscores with spaces, capitalise first letter of each word?), or just something like "AequilibraE Project" as a default

aequilibrae/utils/generate_simwrapper_config.py
```diff
@@ -0,0 +1,503 @@
...
+
+        self._add_to_generated_files("dashboard", output_file)
+
+
+    def _has_links(self):
+        """Checks if project has a network with links"""
+        return True
+
+    def _has_nodes(self):
+        """Checks if project has a network with nodes"""
```
tomstephen: Unlike zones and skims, I think it's a hard requirement for an AequilibraE project to have links, nodes (otherwise it wouldn't have a network!). If you want to check and have it fail early, could interrogate something like [this](https://www.aequilibrae.com/latest/python/useful_links/_autosummary/aequilibrae.project.network.nodes.html).

aequilibrae/utils/simwrapper_panel.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def to_dict(self):
+        """Returns dictionairy representation of the panel"""
+        panel = super().to_dict()
+
+        panel["dataset"] = self.dataset
+
+        return panel
+
+# my = TilePanel(title, dataset)
```
tomstephen: If this is kept as an example on usage, let's add to docstring for `TilePanel` itself. Same goes for TextPanel etc

aequilibrae/utils/spatialite_utils.py
```diff
@@ -51,7 +51,7 @@ def _connect_spatialite(path_to_file: os.PathLike, missing_ok: bool = False):
 
 def load_spatialite_extension(conn: Connection):
     conn.enable_load_extension(True)
-    conn.load_extension("mod_spatialite")
+    conn.load_extension("/opt/homebrew/lib/mod_spatialite")
```
jamiecook: hardcoded

Jake-Moss: Actual fix is in #743

aequilibrae/utils/simwrapper_panel.py
PuttiKoselig: These comments have been updated

aequilibrae/utils/generate_simwrapper_config.py
PuttiKoselig: Have left these for now as empty but can delete if you think they are totally unnecessary

aequilibrae/utils/generate_simwrapper_config.py
PuttiKoselig: Done

aequilibrae/utils/generate_simwrapper_config.py
PuttiKoselig: Completed on line 506

aequilibrae/utils/generate_simwrapper_config.py
PuttiKoselig: resolved via addition of _find_project_title()

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,590 @@
...
+            {"table": "nodes",
+             "geometry": "point",
+             "sqlFilter": "is_centroid=0",
+             "style": regular_node_style},
+        )
+
+        # retun panel inside a list
+        return [panel]
+
+    def _set_default_defaults(self, panel):
```
tomstephen: Two thoughts,

1. If we/someone are setting defaults to the "default" value, we probably want them to act as a debug tool (i.e., we are doing all of the styling ourselves, and the defaults should be a fallback), hence I would recommend picking a really obvious colour we are unlikely to actually want in a visualisation (neon green?) and making the lineWidth/pointRadius really large (1000? Remember units in m)
2. We could implement this as the default value in the panel class itself, either as the default or `defaults_dict=None` then check if None. 

https://github.com/PuttiKoselig/aequilibrae/blob/95c6253d2395cb56d8cf42376e5a14604364b341/aequilibrae/utils/simwrapper/simwrapper_panel.py#L135

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,590 @@
...
+                    {"key": "Node Count",
+                    "value":
+                        {"database": "project_database.sqlite",
+                         "query": "SELECT COUNT(*) FROM nodes"}}]
+
+        panel = TilePanel("Network Size", dataset)
+
+        return [panel]
+
+    def _entire_network_row(self):
```
tomstephen, line 294: Since we are doing centroid & not-centroid nodes, let's do centroid & not-centroid links (i.e. centroid connectors vs all other links)

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,590 @@
...
+        """resturns project details text panel"""
+
+        return [TextPanel(title="title", data="intro")]
+
+    def _stats_rows(self):
+        """returns stats rows panels"""
+        dataset = [{"key": "Link Count",
+                    "value":
+                        {"database": "project_database.sqlite",
+                         "query": "SELECT COUNT(*) FROM links"}},
```
tomstephen: We can add thousands separators using `printf`, so this would become something like `SELECT printf('%,d', COUNT(*)) FROM links`. Also, let's add zone count (either count entries in zones table, or nodes when is_centroid=1) and see the colour palette to monochrome (`colors: monochrome`)

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+        try:
+            folder_name = self.project.project_base_path.name
+            title = folder_name.replace("_", " ").title()
+
+            if not title.strip():
+                raise ValueError
+
+            return title
+
+        except Exception:
```
Jake-Moss: What exception is being caught here? A bare `except Exception` can hide a lot of issues.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+
+    def _find_project_title(self):
+        """Generate  project title from the project folder name, otherwise returns "AequilibraE Project" """
+
+        try:
+            folder_name = self.project.project_base_path.name
+            title = folder_name.replace("_", " ").title()
+
+            if not title.strip():
+                raise ValueError
```
Jake-Moss: What's gone wrong here?

AequilibraE projects also have a model name field. Though it is rarely used.
Image

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+
+    def _categorical_palette(self, n):
+        """Returns n visually distinct colors"""
+
+        base = [
+            "#4C78A8", "#F58518", "#E45756",
+            "#72B7B2", "#54A24B", "#EECA3B",
+            "#B279A2", "#FF9DA6", "#9D755D",
+        ]
+        return base[:n]
```
Jake-Moss: I won't comment on all the colour strings, just noting that we should have some way of utilising a colour palette. 

We might be able to use the matplotlib colormap API https://matplotlib.org/stable/gallery/color/colormap_reference.html

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+        """Builds yaml config for map of entire network"""
+
+        # aequilibrae panel with center and zoom
+        panel = AequilibraEMapPanel(
+            "Entire Network",
+            height=10,
+            width=6,
+            center=self.center,
+            zoom=self.zoom,
+            projection="EPSG:32719",
```
Jake-Moss: Seems model specific

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+            ]
+        )
+
+        # non-centroid connector links
+        panel.add_layer(
+            "links_regular",
+            {
+                "table": "links",
+                "geometry": "line",
+                "sqlFilter": "link_type != 3",
```
Jake-Moss: link types names are not fixed. AequilibraE itself only defines two link type names `centroid_connector` and `default` **if** the network was created using aequilibrae APIs. if it wasn't they could be anything. Which is how I presume the network you've been working with was created. 

We should have some way of specifying this.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+            },
+        )
+
+        # centroid connector links
+        panel.add_layer(
+            "links_centroid_connectors",
+            {
+                "table": "links",
+                "geometry": "line",
+                "sqlFilter": "link_type = 3",
```
Jake-Moss: Same as above, I won't comment on any more of these

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+
+        panel.add_layer(
+            "links",
+            {
+                "table": "links",
+                "geometry": "line",
+                # "sqlFilter": "link_type != 3",
+                "style": capacity_styling,
+            },
+        )
```
Jake-Moss: I'm probably just misunderstanding how this works but it looks like theres a legend for capacities up to 10k, but the data range is set to 1k max? Does capacity get plotted as both width and colour? What about `capacity_ba`? I get the mixed network can be harder to deal with, should we be taking the max of both directions and using that?

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+        """makes scenario comparison map for a network's performance metric'
+
+        :Arguments:
+            **title** (:obj:`str`): panel title
+            **results_table** (:obj:`str`): results table to join to links
+            **metric_column** (:obj:`str`): metric to use for link colouring
+            **legend** (:obj:`list`): legend def for the map
+            **data_range** (:obj:`list`): value range used for colour scale
+            **palette** (:obj:`str`, *Optional*): colour palette to use
+            **width_by_link_type** (:obj:`bool`, *Optional*): vary line width by link type????????? weird
```
Jake-Moss:
```suggestion
            **width_by_link_type** (:obj:`bool`, *Optional*): vary line width by link type
```

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+        return [panel]
+
+    def _scenario_metric_map(
+        self,
+        title,
+        results_table,
+        metric_column,
+        legend,
+        data_range,
+        palette="Temps",
```
Jake-Moss: How does this palette work? Can this be used for my previous comments?

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+        width_by_link_type=True,
+        width_by_metric=None,
```
Jake-Moss: Undocumented argument

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,609 @@
...
+                    1: 20,
+                },
+            }
+
+        if width_by_metric:
+            style["lineWidth"] = {
+                "column": width_by_metric,
+                "dataRange": [0, 500],  # change depending on project/ guess better
+                "widthRange": [10, 250],
+            }
```
Jake-Moss: What happens if neither of these are provided?

aequilibrae/utils/simwrapper/simwrapper_panel.py
```diff
@@ -0,0 +1,325 @@
...
+
+        self.results_database = results_database
+        self.colour_metric = colour_metric
+        self.width_metric = width_metric
+        self.palette = palette
+        self.results_table = results_table
+
+
+        super().set_extra_databases({"results": self.results_database})
+        super().set_legend(self.build_legend())
```
Jake-Moss: Don't need super calls here unless you want to prevent a subclass from overriding the method that gets invoked here.


```suggestion
        self.set_extra_databases({"results": self.results_database})
        self.set_legend(self.build_legend())
```

aequilibrae/utils/simwrapper/simwrapper_panel.py
```diff
@@ -0,0 +1,325 @@
...
+                    "widthRange": [10, 250],
+                }
+            }
+        else:
+            self.width_style = {
+                "lineWidth": 20
+            }
+
+
+    def add_layer(self):
```
Jake-Moss: Overridden method doesn't accept the same arguments as it's super. Perhaps it should be called something else?

aequilibrae/utils/simwrapper/simwrapper_panel.py
```diff
@@ -0,0 +1,325 @@
...
+                },
+                "style": [self.colour_style, self.width_style],
+            },
+        )
+
+    def to_dict(self):
+        """Returns dictionary representation of the panel"""
+        panel = super().to_dict()
+
+        return panel
```
Jake-Moss: On methods like this you can just omit them and it'll call the super anyway.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,607 @@
...
+        # skip if no convergence data
+        if csv_path is None:
+            return None
+
+        vega_spec = self._write_convergence_vega_spec(csv_path)
+
+        # panel wrapper
+        panel = ConvergencePanel(
+            title="Assignment Convergence",
+            config="simwrapper/simwrapper_data/" + vega_spec,
```
Jake-Moss:
```suggestion
            config="simwrapper_data/" + vega_spec,
```

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
```diff
@@ -0,0 +1,607 @@
...
+from aequilibrae.utils.simwrapper.simwrapper_utils import get_project_center, get_project_zoom
+
+
+class SimwrapperConfigGenerator:
+    """
+    Generate simwrapper ready .yaml file from an AequilibraE Project with
+    minimal manual work.
+    """
+
+    def __init__(self, project, output_dir="simwrapper"):
```
Jake-Moss: When testing this it wasn't clear to me that the simwrapper output directory **must** be a subdirectory of the project, and that supplying another directory should only be used to renamed the sub-directory.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
Jake-Moss: This breaks when the output directory argument is specified.

Jake-Moss (approved):
Nice work!

It would also be nice to have this available as a pyproject cli script such that it can be used without creating your own script given the typically usage is very simple. https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts

```python
from aequilibrae import Project
from aequilibrae.utils.simwrapper.generate_simwrapper_config import SimwrapperConfigGenerator

p = Project()
p.open("model path")

s = SimwrapperConfigGenerator(p)
s.write_yamls()
```

Jake-Moss:
Please also make sure you are using `ruff format` and `ruff check`

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
PuttiKoselig: Good point, yes, but SimWrapper requires that it is within the project folder that you open. Will add to documentation.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
PuttiKoselig: Intention would be to use something like that for the new, and if we don't have it format the project folder name for a guess at the title.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
PuttiKoselig: Intention was to default to "AequilibraE Project" if folder name revealed nothing

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
PuttiKoselig: This was intended as a planned way to use distinct colors in certain maps. Palettes introduced later.

aequilibrae/utils/simwrapper/generate_simwrapper_config.py
PuttiKoselig: Nothing good, breaks. Needs default, needs to handle ranges like the legends.

Jake-Moss:
Thanks for all your work on this @PuttiKoselig and @tomstephen !

PR #746, opened 2026-01-29
Pedro/separate pt map match

pedrocamargo:


aequilibrae/transit/lib_gtfs.py
```diff
@@ -48,12 +50,10 @@ def __init__(self, network, agency_identifier, file_path, day="", description=""
...
             self.gtfs_data = GTFSReader(conn)
         self.srid = get_srid()
-        self.transformer = None
-        self.wgs84 = pyproj.Proj("epsg:4326")
+        self.mm_transformer: Transformer
         self.trip_by_service = {}
         self.patterns = {}
-        self.graphs = {}
-        self.transformer = Transformer.from_crs("epsg:4326", f"epsg:{self.srid}", always_xy=False)
+        self.map_matchers = {}  # Dict[str, RouteMapMatcher]
```
Jake-Moss:
```suggestion
        self.map_matchers: dict[str, RouteMapMatcher] = {}
```

aequilibrae/transit/lib_gtfs.py
```diff
@@ -423,25 +420,37 @@ def _get_fare_attributes_by_date(self, service_date: str) -> dict:
...
+
+            if link_gdf.shape[0] == 0 or nodes_gdf.shape[0] == 0 or stops_gdf.shape[0] == 0:
                 continue
-            self.graphs[mode] = graph
+            rmm = RouteMapMatcher(link_gdf, nodes_gdf, stops_gdf)  # type: ignore
+            rmm.initialize_graph()
+            self.map_matchers[pt_mode] = rmm
+
+            self.mm_transformer = Transformer.from_crs(self.srid, self.map_matchers[pt_mode].crs, always_xy=True)
+            self.mm_transform_rev = Transformer.from_crs(self.map_matchers[pt_mode].crs, self.srid, always_xy=True)
```
Jake-Moss: This doesn't look right, the attribute will be replaced each iteration. If there's `pt_modes` not already in `self.map_matchers` then these won't be set at all

aequilibrae/transit/route_map_matcher.py
```diff
@@ -0,0 +1,348 @@
...
+from aequilibrae.utils.geo_utils import metre_crs_for_gdf
+from aequilibrae.utils.interface.worker_thread import WorkerThread
+
+DEAD_END_RUN = 40
+
+
+class RouteMapMatcher:
+    def __init__(self, link_gdf: gpd.GeoDataFrame, nodes_gdf: gpd.GeoDataFrame, stops_gdf: gpd.GeoDataFrame,
+                 distance_to_project=50):
+        WorkerThread.__init__(self, None)
```
Jake-Moss:
```suggestion
class RouteMapMatcher(WorkerThread):
    def __init__(self, link_gdf: gpd.GeoDataFrame, nodes_gdf: gpd.GeoDataFrame, stops_gdf: gpd.GeoDataFrame,
                 distance_to_project=50):
        super().__init__(self, None)
```

This calls the init method of another unrelated class on this object. I think this was supposed to inherit from `WorkerThread`

aequilibrae/transit/route_map_matcher.py
```diff
@@ -0,0 +1,348 @@
...
+        access_links = self.graph.network[self.graph.network.a_node.isin(route_stops.stop_id.values)]
+        path_links, path_directions = [], []
+
+        is_first = True
+        for i in range(1, route_stops.shape[0]):
+            next_stop = int(route_stops.stop_id[i])
+            is_not_last = i < route_stops.shape[0] - 1
+
+            # Get the next stop for look-ahead path estimation (if not at the end)
+            following_stop = int(route_stops.stop_id[i + 1]) if is_not_last else None
```
Jake-Moss: Does `route_stops` definitely have an index that lines up with `[1, route_stops.shape[0])`? I can see an inner merge on line 132 that might mess with the index? If not should be be done with `route_stops.stop_id.iat[i]`?

aequilibrae/transit/route_map_matcher.py
```diff
@@ -0,0 +1,348 @@
...
+            }
+        )
+
+        # Map internal link IDs to original network IDs
+        df = result.merge(corresp, on="link_id", how="left")
+        df.sort_values(by=["sequence"], inplace=True)
+
+        # Remove consecutive links with same original_id and direction
+        # (these are internal graph subdivisions of the same physical link)
+        df = df[(df.original_id.shift(-1) != df.original_id) | (df.direction.shift(-1) != df.direction)]
```
Jake-Moss: Shift without a fill value on numeric types will insert `np.nan` and cast the column to float, with large enough integers they will become equal (e.g. `float(2**60 + 1) == float(2**60)`)

aequilibrae/transit/route_map_matcher.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # Filter out isolated short segments (likely noise or dead-end detours)
+        # Keep a link if it differs from both neighbors OR if it's long enough
+        crit_differs_prev = df.original_id.shift(1) != df.original_id
+        crit_differs_next = df.original_id.shift(-1) != df.original_id
```
Jake-Moss: Same as above

aequilibrae/transit/route_map_matcher.py
```diff
@@ -0,0 +1,348 @@
...
+        # Eliminate back-and-forth patterns on the same link
+        # e.g., [A, B, A] -> [A] when all three have the same absolute link_id
+        has_issues = True
+        while has_issues:
+            has_issues = False
+            for i in range(0, df.shape[0] - 2):
+                # Check if three consecutive rows are all the same link
+                if df.loc[i: i + 2, "original_id"].unique().shape[0] == 1:
+                    df.drop(index=[i, i + 1], inplace=True)
+                    df.reset_index(drop=True, inplace=True)
```
Jake-Moss: I can see this repeated dropping tanking pandas performance, is it possible to mark these as "bad" then drop them all at once later?

`.nunique()` might be faster than a `unique().shape[0]`, not sure though.

The index looks like it will be right here but things like `iloc` and `iat` will always be faster than a label based lookup.

aequilibrae/transit/functions/breaking_links_for_stop_access.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # Calculate distance along line (linear reference)
+    line_geoms = df_result["geometry_line"].values
+    point_geoms = df_result["geometry"].values
+
+    # Use vectorized project if available, otherwise apply
+    dist_along = shapely.line_locate_point(line_geoms, point_geoms)
+
+    df_result["dist_along"] = dist_along
+    df_result["line_length"] = df_result["geometry_line"].apply(lambda x: x.length)
```
Jake-Moss: This is a series with object dtype rather than geometry? If it's a GeoSeries with dtype `geometry` you can still use `.length`.

aequilibrae/transit/functions/breaking_links_for_stop_access.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # --- Step 2: Prepare Splits ---
+
+    # Filter out splits that are effectively at the start or end of the link
+    # 1 metre is effectively "right there"
+    epsilon = 1
+    valid_splits = df_result[
+        (df_result["dist_along"] > epsilon) &
+        (df_result["dist_along"] < (df_result["line_length"] - epsilon))
+        ].copy()
```
Jake-Moss:
```suggestion
    valid_splits = df_result[
        (df_result["dist_along"] > epsilon) &
        (df_result["dist_along"] < (df_result["line_length"] - epsilon))
    ].copy()
```

Jake-Moss (approved):
Something looks odd with the CI. I'll retrigger it and see if we see the same failure

PR #752, opened 2026-02-11
Fixes subclassing

pedrocamargo:
This was causing issues when inside QGIS


Jake-Moss (approved):
Yep that’ll do it, whoops

PR #753, opened 2026-02-12
Add Akcelik VDF

Jake-Moss:
Note: the parameter 'tau' value has typical `8` factor absorbed into it. Users should supply `8 * tau` to match other common usages.

image

Which matches "Travel time functions for transport planning purposes: Davidson's function, its time-dependent form and an alternative travel time function" pretty well
image

image

And our derivative matches a central finite differences.
image


pedrocamargo (approved):
We are still missing user documentation, but that can come in a future PR

Issue #756, opened 2026-02-13
Repository gardening


Jake-Moss:
Would be nice to clean the repo structure up a bit. That would be removing unused files, migrating configs to pyproject.toml, move to src-layout instead of flat. De-duplicate the workflows, fix that annoying license warning etc.

PR #757, opened 2026-02-13
Repository gardening

Jake-Moss:
#756

- [X] Removing unused files.
- [X] Migrating configs to pyproject.toml.
- [ ] ~Move to src-layout.~
- [ ] ~De-duplicate the workflows.~
- [X] Fix that annoying license warning.
- [X] Fix all the warnings that removing the ruff-ignores brought back. A lot of them are just imports that need to be rewritten to "import x as x" to prevent "import unused".


pedrocamargo (approved):
It looks mostly good, but I am not sure the move to *.cython.* is a positive one. 
While it is perfectly fine for internal tools, it feels clunky for things like the ipf_core, which is arguably accessed directly by users.

The solution may be to have all the Cython code in the same place (e.g. aequilibrae.core)?

PR #762, opened 2026-02-20
Fix missing tabs display in docs

tomstephen:
A [recent commit](https://github.com/AequilibraE/aequilibrae/commit/0ce291315037b5b5c6c26eb10029d8c124f9a850) disabled the `sphinx-tabs` extension when the sphinx version is too high, as it has been deprecated. This breaks rendering in the docs. For example, see the [expected output here](https://www.aequilibrae.com/docs/python/v1.4.1/traffic_assignment/traffic_assignment_validation.html) and the [broken output here](https://www.aequilibrae.com/latest/python/traffic_assignment/traffic_assignment_validation.html). 

`sphinx-design`, which is already being used in our docs, implements a tab style as well, so this PR removes `sphinx-tabs` and adjusts the docs to use `sphinx-design` instead.

The new style is (hopefully acceptably) different. See below for an example.

Screenshot 2026-02-20 at 5 03 47 pm

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -66,15 +66,11 @@ extensions = [
     "sphinx_design",
     "sphinx_copybutton",
     "sphinx_git",
-    "sphinx_tabs.tabs",
+    "sphinx_design",
     "sphinx_subfigure",
 ]
 
-if int(sphinx.__version__.split(".")[0]) >= 9:
-    # sphinx-tabs fails with Sphinx 9+ (KeyError: backrefs), so omit it.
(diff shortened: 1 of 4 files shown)


pedrocamargo (approved):

pedrocamargo:
Nice !!!

Jake-Moss:
Do we want a patch release to fix the docs page?

pedrocamargo:
> Do we want a patch release to fix the docs page?
I can overwrite things directly on AWS...

PR #764, opened 2026-03-02
Fixes SimWrapper

pedrocamargo:
Without the __init__.py, the entire folder is ignbored during packaging.

@Jake-Moss , that "src" change will save from these things too!!!


jamiecook (approved):

Jake-Moss:
Ugh yep that'll do it @pedrocamargo . sdist and wheels look good now. Thanks. The src-layout would deifnitely help with this by allowing use to test the wheels directly without Python accidentality picking up a local clone/editable install.

Related to that, I really don't like the "we'll find all your files don't worry" approach to the packaging. [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/) helps with this but it adds all files tracked by git, so you need to exclude anything you don't want. [hatch](https://github.com/pypa/hatch) is the now recommended back-end build system, in combination with [build](https://github.com/pypa/build) as the front end, and can do a similar approach but needs an extension [hatch-cython](https://github.com/joshua-auchincloss/hatch-cython) for Cython support. There is also [scikit-build-core](https://github.com/scikit-build/scikit-build-core) for a CMake based build system (way over built for our needs but is neat).

The one I'm most interested is [meson-python](https://mesonbuild.com/meson-python/) which uses the meson build system (like CMake but faster and more user friendly). It's got good Cython support and some neat auto-recompile-on-import features for "editable installs". It also allows multiple build configurations to exist at once (multiple builds for different Python versions installed at once without conflicting). But there was some issues with "[data files](https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html#data-files)" I was waiting to be resolved. They're fixed now in Python 3.10 (stdlib was incorrect) and it only required minimal changes.

pedrocamargo:
> Ugh yep that'll do it @pedrocamargo . sdist and wheels look good now. Thanks. The src-layout would deifnitely help with this by allowing use to test the wheels directly without Python accidentality picking up a local clone/editable install.
> 
> Related to that, I really don't like the "we'll find all your files don't worry" approach to the packaging. [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/) helps with this but it adds all files tracked by git, so you need to exclude anything you don't want. [hatch](https://github.com/pypa/hatch) is the now recommended back-end build system, in combination with [build](https://github.com/pypa/build) as the front end, and can do a similar approach but needs an extension [hatch-cython](https://github.com/joshua-auchincloss/hatch-cython) for Cython support. There is also [scikit-build-core](https://github.com/scikit-build/scikit-build-core) for a CMake based build system (way over built for our needs but is neat).
> 
> The one I'm most interested is [meson-python](https://mesonbuild.com/meson-python/) which uses the meson build system (like CMake but faster and more user friendly). It's got good Cython support and some neat auto-recompile-on-import features for "editable installs". It also allows multiple build configurations to exist at once (multiple builds for different Python versions installed at once without conflicting). But there was some issues with "[data files](https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html#data-files)" I was waiting to be resolved. They're fixed now in Python 3.10 (stdlib was incorrect) and it only required minimal changes.

I am happy for you to lead the way and choose the solution we will adopt, @Jake-Moss !  I guess it makes sense to have this as part of the minimum changes to 2.0

PR #780, opened 2026-04-20
type checking

EvelynBond:


src/aequilibrae/paths/graph.py
```diff
@@ -9,6 +9,7 @@
 from typing import List, Optional, Tuple, Union
 
 import numpy as np
+import numpy_typing as npt
```
pedrocamargo, line 12: Do we want a new dependency?

src/aequilibrae/paths/graph.py
```diff
@@ -459,17 +460,18 @@ def set_skimming(self, skim_fields: list) -> None:
         if k:
             raise ValueError("At least one of the skim fields does not exist in the graph: {}".format(",".join(k)))
 
-        if self.centroids.shape[0]:
-            self.compact_skims = np.zeros((self.compact_num_links + 1, len(skim_fields) + 1), self.__float_type)
+        if self.centroids is not None:
```
pedrocamargo: A question for @Jake-Moss . Do we ever want to centroids to be None? Or an empty array would mean no centroids?

pyproject.toml
```diff
@@ -31,7 +31,7 @@ readme = "README.md"
 requires-python = ">=3.10" # match classifiers above and ruff version below
 
 dependencies = [
-  "numpy",
+  "numpy>=1.21",
```
pedrocamargo: Let's require Numpy >=2.0

src/aequilibrae/paths/graph.py
jamiecook: It's weird atm as there is alredy this in prepare_graph:

Image

Seems like calling prepare graph once is fine with no centroids (None) but calling it a second time would error as they are now empty set....

pyproject.toml
jamiecook: done

jamiecook (approved):

PR #785, opened 2026-05-12
Merge with my type checking

TylerP132:


src/aequilibrae/project/network/safe_class.py
```diff
@@ -1,8 +1,18 @@
+from __future__ import annotations
```
Jake-Moss, line 1: Are we using the delayed evaluation of annotations here?

https://docs.python.org/3/library/__future__.html#future__.annotations

src/aequilibrae/transit/functions/breaking_links_for_stop_access.py
```diff
@@ -167,9 +166,9 @@ def split_links_at_stops(
     broken_links = pd.concat([untouched_links, new_segments_gdf], ignore_index=True)
 
     broken_links = broken_links.assign(original_id=broken_links.link_id)
-    broken_links = broken_links.assign(link_id=np.zeros(broken_links.shape[0]) + 1)
+    broken_links = gpd.GeoDataFrame(broken_links.assign(link_id=np.zeros(broken_links.shape[0]) + 1))
```
Jake-Moss: We should also assign the crs here as well like line 172 if this is now a GeoDataFrame

src/aequilibrae/transit/route_map_matcher.py
```diff
@@ -23,28 +23,32 @@ def __init__(
         link_gdf: gpd.GeoDataFrame,
         nodes_gdf: gpd.GeoDataFrame,
         stops_gdf: gpd.GeoDataFrame,
-        distance_to_project=50,
+        distance_to_project: int | float = 50,
```
Jake-Moss, line 26: `float` is enough here, there's a really odd special case in Python where a `float` annotation is actually `float | int`.

https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex

This discourse post lists a bunch of issues with this but it exists and isn't going away.
https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018/1

Jake-Moss (approved):
Nice work. Thanks for all these, I'll fix up the geopandas crs comment but then I'm happy to merge. Other comments are just FYIs.

src/aequilibrae/project/network/safe_class.py
TylerP132: Yes, since when running `Project` is not defined and it is used in the constructor of `SafeClass`.

PR #794, opened 2026-05-28
Fix units in Akcelik VDF

Jake-Moss:
Time in the Akcelik VDF is actually time per distance, this implements the expanded version to avoid a division and multiplication.

### DIFF
--- a/tests/aeq/paths/test_akcelik.py
+++ b/tests/aeq/paths/test_akcelik.py
@@ -39,14 +39,15 @@ def test_akcelik_function():
 
     np.testing.assert_allclose(should_be, congested_times, err_msg="Akcelik is wrong")
 
+    length = np.ones(link_flows.shape[0])
     # Let's check the derivative for sections of the curve
     dx = 0.00000001
     for i in range(1, 11):
         link_flows.fill(1 * 0.2 * i)
         link_flows += np.arange(11) * dx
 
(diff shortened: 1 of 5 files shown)


tests/aeq/paths/test_akcelik.py
```diff
@@ -39,14 +39,15 @@ def test_akcelik_function():
...
 
+    length = np.ones(link_flows.shape[0])
     # Let's check the derivative for sections of the curve
     dx = 0.00000001
     for i in range(1, 11):
         link_flows.fill(1 * 0.2 * i)
         link_flows += np.arange(11) * dx
 
-        akcelik(congested_times, link_flows, capacity, fftime, alpha, tau, cores)
-        delta_akcelik(delta, link_flows, capacity, fftime, alpha, tau, cores)
```
jamiecook, line 49: Is it worth adding some actual tests on the expected values from this? I note that this test as it stands didn't catch the fact that the values coming out previoulsy were entirely wrong

jamiecook (approved):

tests/aeq/paths/test_akcelik.py
pedrocamargo: It did have tests. We do not have an external reference, though.  So the implementation of the function and the tests were both wrong

Jake-Moss:
Merged into 2.0 in e16020a9

PR #805, opened 2026-07-02
Sorting optimization to fix windows qsort slowdown

nick-fournier:
Sorry to throw this PR out of left field, but this one of those satisfyingly tiny fixes with a massive impact (like 300x speedup) for those unfortunate souls that are Windows bound. It now outperforms legacy commercial software that rhymes with tube.

I found that when I ran transit assignment it was absurdly slow, which was surprising because I know this to be fast. Even weirder was it was faster on the same machine when run via WSL. So with a little help from the robots for profiling and a few hours digging into the absurd details of sorting algorithms, it turns out it was a Windows specific bug (of course).

In the robot's words:
> For every destination, the code sorted the entire edge array (say 500k edges), even though only the few hundred hyperpath edges (h_a_vec == 1) needed ordering. All non-hyperpath edges got the same placeholder key 1.0. MSVC's qsort is a quicksort, which degrades toward O(n²) on input that's almost all ties — glibc's is a mergesort, which doesn't, hence Windows-only pain.

> Fix (hyperpath.pyx): Compact the hyperpath edges to the front of the array first, then sort only those. Sorting ~200 distinct keys instead of ~500k mostly-tied ones — the pathology can't trigger and the work shrinks by orders of magnitude.

> Bonus fix (public_transport.pyx): The demand table was also being fully re-scanned once per destination to find its rows. Now it's sorted by destination once up front, so each destination's demand is a contiguous slice found via demand_indptr — O(1) lookup instead of a full scan. The stable sort keeps results identical.

> Sorting context: 

> * glibc (Linux): implements qsort as a mergesort (falling back to quicksort only if memory allocation fails). Mergesort is O(n log n) no matter what the input looks like — ties are irrelevant.
> * MSVC (Windows): implements qsort as an actual quicksort. Classic quicksort has a known worst case: when the array is dominated by equal keys, partitioning stops making progress (every element compares equal to the pivot, so the splits are maximally unbalanced) and runtime degrades toward O(n²).

### DIFF
--- a/aequilibrae/paths/cython/hyperpath.pyx
+++ b/aequilibrae/paths/cython/hyperpath.pyx
@@ -190,6 +190,7 @@ cdef void compute_SF_in_parallel(
     uint32_t[::1] head_view,
     uint32_t[:] d_vert_ids_view,  # destination vertices
     uint32_t[:] destination_vertex_indices_view,
+    uint32_t[:] demand_indptr_view,  # start of each destination's slice in the destination-sorted demand arrays
     uint32_t[:] rest_of_destinations_view,  # Used when skimming to lookup the rest of the destinations to skim from
     uint32_t[::1] o_vert_ids_view,  # origin vertices
     double[::1] demand_vls_view,  # volume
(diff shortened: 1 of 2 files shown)


Jake-Moss:
Hi @nick-fournier , thank you for this PR!

I was been able to reproduce the issue, and your PR does fix it. I was able to see a reduction from 3-4mins per assignment down to ~3s on windows, and a slight improvement on WSL from ~3s to ~2s. This was tested on a network with 4616 zones, random demand matrix filled to 20% with ~70k transit links.

I had some issues pushing to the PR branch so I've another with your commit + some additional changes, see #806 .

PR #806, opened 2026-07-03
Sorting optimization to fix windows qsort slowdown

Jake-Moss:
Closes #805 

This contains @nick-fournier's commit, plus some additional changes, I've
- moved the allocation out of a loop and into the initial scratch allocations,
- removed the assignment of `origin_column`, `destination_column`, and `demand_column` to the `HyperpathGenerating` object, we shouldn't be storing them and mutating them. These are pretty far down in the library so I'm not worried that removing these will break peoples code, however, anyone or @nick-fournier depends on them we can restore them, and
- made the comparator function passed to qsort to return 0 when objects are actually equal. This brings the runtime down from 3-4mins to ~5s alone, before any other changes. The reduced sorting changes still provide another ~2s of runtime improvement.

Thank you brings this to our attention @nick-forunier, this code was definitely profiled and tested by myself on linux-only systems.


Jake-Moss:
Merged into 2.0 in 0216b5f9

PR #809, opened 2026-07-10
Allow override to_parquet arguments in route choice route set saving

Jake-Moss:
Allows the user to override all arguments used to save the route choice sets. A user ran into this when saving choice sets for more origins than the default `max_partitions`.

### DIFF
--- a/aequilibrae/paths/route_choice.py
+++ b/aequilibrae/paths/route_choice.py
@@ -287,6 +287,7 @@ class RouteChoice:
             path_size_logit=bool(demand),
             cores=self.cores,
             where=str(self.where) if self.where is not None else None,
+            to_parquet_kwargs=self.to_parquet_kwargs,
             sl_link_loading=self.sl_link_loading,
             **self.parameters,
         )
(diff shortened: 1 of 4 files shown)


aequilibrae/paths/route_choice.py
```diff
@@ -177,7 +178,7 @@ def set_cores(self, cores: int) -> None:
         """
         self.cores = cores
 
-    def set_save_routes(self, where: Optional[str] = None) -> None:
+    def set_save_routes(self, where: Optional[str] = None, **to_parquet_kwargs) -> None:
```
jamiecook: I'd consider making the user facing set of kwargs also contained to a named dict

def set_save_routes(self, where: Optional[str] = None, to_parquet_kwargs: dict) -> None:

Rather than just collecting together all keywords that are passed here into a dict... this would mean that users would have to call it 

set_save_routes(to_parquet_kwargs={"foo": 7})

but I think it's nicer to be more explicit - there are some of the passthrough kwargs in matplotlib that do my head in.

jamiecook (approved):

Jake-Moss:
Merged into 2.0 in eb1fa0d8

PR #812, opened 2026-07-19
Fix for recurrent use of Spatialite on Windows

pedrocamargo:


Jake-Moss:
I can see this 1088 TLS index limit documented at https://learn.microsoft.com/en-us/windows/win32/procthread/thread-local-storage , however, I can't see how this fixes it? When the SQLite extension is loaded it's `sqlite3_modspatialite_init` function is called, which presumably uses uses the TLSAlloc function and is assigned a new TLS index. Every time we load the extension it will do this again.

I haven't had a chance to repo this, nor to check this actually fixes it but I will later today.

I think the actual fix here might be to not constantly load and reload the extension. I recall we had issues with the connection in QGIS which is why we constantly close the connection but this shouldn't be required.

> (5) Can multiple applications or multiple instances of the same application access a single database file at the same time?
>
>    Multiple processes can have the same database open at the same time. Multiple processes can be doing a SELECT at the same time. But only one process can be making changes to the database at any moment in time, however.

pedrocamargo:
> I can see this 1088 TLS index limit documented at https://learn.microsoft.com/en-us/windows/win32/procthread/thread-local-storage , however, I can't see how this fixes it? When the SQLite extension is loaded it's `sqlite3_modspatialite_init` function is called, which presumably uses uses the TLSAlloc function and is assigned a new TLS index. Every time we load the extension it will do this again.
> 
> I haven't had a chance to repo this, nor to check this actually fixes it but I will later today.
 
I don't quite understand why this works either, but the fact is that tests no longer fail randomly after too many loads/unloads of Spatialite
 
> I think the actual fix here might be to not constantly load and reload the extension. I recall we had issues with the connection in QGIS which is why we constantly close the connection but this shouldn't be required.
> 
> > (5) Can multiple applications or multiple instances of the same application access a single database file at the same time?
> > Multiple processes can have the same database open at the same time. Multiple processes can be doing a SELECT at the same time. But only one process can be making changes to the database at any moment in time, however.

SQLite allows for multiple parallel reads, but things get pretty hairy when doing changes, so I think it is a good practice to open and close whenever we need to use it.

In practice, this happens very little (accessing the database), and ends up being an issue only during testing.

Jake-Moss:
I can repo the crash and verify that opening the dll manually with ctypes prevents it. I think this is to keep the reference count above one and thus prevents unloading the library.

> SQLite allows for multiple parallel reads, but things get pretty hairy when doing changes, so I think it is a good practice to open and close whenever we need to use it.

I don't think this is the case? SQLite absolutely supports concurrent reads and write (singular) when configured correctly (WAL, busy timeout set)

> I don't quite understand why this works either, but the fact is that tests no longer fail randomly after too many loads/unloads of Spatialite

Do you think this is part of the CI failures? I haven't ever run into this locally, and CI also fails on other OSs as well.

Jake-Moss (approved):

Jake-Moss:
Merged into 2.0 in 5d854668

PR #813, opened 2026-07-20
Custom Volume Delay Functions (VDF) 

TylerP132:
…cumentation and tests, and sketched api design for future implementation.


src/aequilibrae/paths/traffic_assignment.py
```diff
@@ -323,14 +320,21 @@ def __check_attributes(self, instance, value):
             return False, value, f"TrafficAssignment class does not have property {instance}"
         return True, value, ""
 
-    def set_vdf(self, vdf_function: str) -> None:
+    def set_vdf(
+        self,
+        name: str,
+        func: Callable,
+        spec: dict,
+        d_func: Callable | None = None,  # central difference
```
jamiecook: just pass in a VDF object?

src/aequilibrae/paths/vdf.py
```diff
@@ -11,17 +12,176 @@
...
+
+            elif "functional_form" in entry:
+                # Compiling an arbitrary math expression string into a callable (with a
+                # derivative, correct signature, and cores-based parallelization matching
+                # every other VDF here) isn't implemented yet - needs its own
+                # parser/compiler (e.g. sympy.lambdify + autodiff for the derivative).
+                raise NotImplementedError(
+                    f"VDF '{name}' defines a custom 'functional_form', but compiling "
+                    "arbitrary VDF expressions into callables isn't implemented yet."
+                )
```
Jake-Moss: Lets not bother with a custom parser/compiler or auto-differentiation. We'll use numexpr for the function generation, it's widely used. And then just a finite central difference if a derivative isn't supplied. We accept arbitrary python functions for anything else more advanced.

src/aequilibrae/paths/vdf.py
```diff
@@ -11,17 +12,176 @@
...
+                bounds: [0, .inf]
+        """
+        vdf_data = dict(vdf_data)  # don't mutate the caller's dict
+        self.default = vdf_data.pop("default", None)
+
+        for name, entry in vdf_data.items():
+            if entry is None:
+                continue
+
+            if "function" in entry:
```
Jake-Moss, line 104:
```suggestion
            if "function" in entry and "functional_form" not in entry:
```

src/aequilibrae/paths/vdf.py
```diff
@@ -11,17 +12,176 @@
...
+                function_name = str(entry["function"]).lower()
+                if function_name not in FUNCTION_MAP:
+                    raise ValueError(
+                        f"VDF '{name}' references unknown preset function '{entry['function']}'. "
+                        f"Available presets are: {', '.join(FUNCTION_MAP.keys())}."
+                    )
+                func, derivative = FUNCTION_MAP[function_name]
+                self.add_vdf(name, func, entry["spec"], derivative, override_existing=True)
+
+            elif "functional_form" in entry:
```
Jake-Moss, line 114:
```suggestion
            elif "functional_form" in entry and "function" not in entry:
```

Jake-Moss:
Looking good so far

src/aequilibrae/paths/vdf.py
```diff
@@ -11,17 +12,176 @@
...
+        # project.add_vdf(name="bpr_tyler", function="bpr", spec=bpr_spec)
+        # have wrapper in project that passes all arguments here
+        if not override_existing and name in self.vdfs:
+            raise ValueError(f"A volume delay function of name {name} is already stored.")
+        new_vdf = VDF(name, function, spec, derivative)
+        self.vdfs[name] = new_vdf
+
+    def get_vdf(self, name):
+        name_lower = name.lower()
+        if name_lower in FUNCTION_MAP:
```
jamiecook, line 165: maybe we just .lower everything and make this case insensitive (i.e. can't have bar_tyler and bar_Tyler as different things)

src/aequilibrae/paths/vdf.py
jamiecook: question... should we be using a numexpr evaluation even for our predefined functions? Currently we have our own Cython implementations - there is a non-trivial chance that numexpr might actually be faster right? Then we'd get to delete code!

Jake-Moss: It's certainly possible, we'll have to test it!

src/aequilibrae/paths/vdf.py
Jake-Moss: I've been taking the opposite stance and removing all case modifications on our end. I've run into a number of cases where code written with a case insensitive file system in mind modifies the case, then breaks on a case sensitive file system. While there's no file IO here I think we should be consistent, and the easiest way to be consistent is to not modify the case.

jamiecook: I'm generally open to that point of view... but I think that when we fully control the storage layer (i.e. not delegating to the os/filesystem) we can be slightly more accommodating of user "mistyping"... If we make it an error to add a duplicate vdf then we can guarantee this won't be an issue at runtime.

src/aequilibrae/paths/vdf.py
jamiecook:
```
Cores Sweep Summary
cores   cython_best(s)  cython_rel   pandas_best(s)  pandas_rel   numexpr_best(s)  numexpr_rel
----------------------------------------------------------------------------------------------
    1         0.007734       1.000x         0.012526       1.620x         0.002407       0.311x
    2         0.006007       0.777x         0.011963       1.992x         0.002076       0.346x
    3         0.004156       0.537x         0.012299       2.959x         0.002129       0.512x
    4         0.003155       0.408x         0.012349       3.914x         0.002098       0.665x
    5         0.004053       0.524x         0.012079       2.980x         0.002121       0.523x
    6         0.003479       0.450x         0.012352       3.550x         0.002074       0.596x
    7         0.003031       0.392x         0.012053       3.977x         0.002110       0.696x
    8         0.002784       0.360x         0.012333       4.431x         0.002135       0.767x
    9         0.002917       0.377x         0.012284       4.211x         0.002077       0.712x
   10         0.002704       0.350x         0.012391       4.582x         0.002111       0.781x
   11         0.002796       0.362x         0.012479       4.462x         0.002106       0.753x
   12         0.002892       0.374x         0.012468       4.311x         0.002074       0.717x
   13         0.003003       0.388x         0.012476       4.154x         0.002086       0.694x
   14         0.002815       0.364x         0.012257       4.354x         0.002081       0.739x
   15         0.002661       0.344x         0.012407       4.662x         0.002129       0.800x
   16         0.002545       0.329x         0.012234       4.807x         0.002059       0.809x
```

jamiecook: so even with a good core count - numexpr is still 20% faster than our Cython implementation of bar

TylerP132:
Making and using custom VDFs works, but the utils around it and documentation are not done. 

What is implemented:
- using preset VDFs
- defining new VDFs by supplying the function and its derivative as callables
- defining a VDF by a string using NumExpr
- VDFs are input from the parameters.yml file, either preset ones or defined by a string interpreted by NumExpr
- some tests for these things

Functionality not implemented yet:
- Check that values in the graph satisfy the bounds of the parameters of the VDF. This was done when setting the VDF parameters in `TrafficAssignment`, and it was previously hard coded for all the preset VDFs. This was replaced by the `_set_vdf_link_attributes` function but it doesn't check that the values are within the bounds
- an option to not supply a derivative, and just use finite difference to estimate the derivative
- plotting all VDFs in a VDFsManager class on the same plot for comparison
- a function that checks that a VDF is reasonable, for example it being positive and strictly increasing
- fix preset conical VDF to adhere to its [definition](http://www.spiess.ch/emme2/conic/conic.html)

The test I made for plotting a VDF just saves the plots in the same folder the test is in, and it probably shouldn't do that.

Previous documentation is wrong and I've fixed some of it. There should also probably be examples of making, plotting and checking custom VDFs etc.

Jake-Moss:
Hey @TylerP132 , had a good chat with Pedro this afternoon about this PR. Overall it's looking good, the yaml spec in particular is going to be quite useful.

Heres some notes from our conversation:
- [x] When using a built in VDF as a custom one (i.e. spec has `function: bpr`), the provided spec is ignored in favour of the built in one. I think we should either disallow it, merge them, or require it be complete.
- [x] We should revert the changes that made `capacity` a non-special case parameter. Pedro pointed out that when we allow multiple VDFs, we should not allow using different capacity fields between VDFs. It's very unlikely to be intended given capacity in AequilibraE is specified in PCE/period. While I think allowing multiple VDFs is equivalent I think it reduces a large footgun.
- [x] We should rethink the "adding" API for the VDFs. To us, "adding" implies some sort of permanent addition when it acts on a "project". Conversely it's not clear that it's an ephemeral operation. I think we should move to using the `VDF` constructor itself for creating runtime VDFs, with a special constructor for built in ones. Then the VDF manager object can move to become a more parameters.yml and project focused object. Allowing reading and perhaps serialisation of individual VDF objects.
- [x] The plotting shouldn't mutate global state (`rcParams`)
- [x] Plotting should return the plot object to play better with notebooks and not save it anywhere, that's the users concern.

On the documentation:
- [ ] The use of `fftime` and `link_flows` as reserved parameter names needs to be documented ~~and enforced. You shouldn't be able to provide another parameter of the same name. It will break atm but I think it should be checked earlier~~. I'm also leaning to renaming `fftime` to `ff_time`.
- [ ] Should we be very explicit about what checks we have implemented, citing the paper and detailing how we are checking (cause it's not analytic).
- [ ] We should use something other than the quadratic and "bpr_tyler" in the examples, I've got a fairly custom one we can use from another project.
- [ ] In the docs we should have a small gallery to showcase "here's some of the things you can do" and include some interesting examples as a bit of novelty and to demonstrate it's a powerful API.

Jake-Moss:
Thanks for this PR @TylerP132 , I'm going to merge this now in order to avoid creating a lot of merge conflicts for you with the other open PRs. My understanding is that it's largely just docs that we are missing now. I'm quite liking the interface for this.

PR #815, opened 2026-07-24
Remove network requests in tests

Jake-Moss:


Jake-Moss:
We got hit with a couple 429s today in the CI, it's been fine the past after a cool off period but I figured it would be good to remove the rng that we currently have to determine if the tests run or not.

This does give up testing against an actual server and real response, but the OSM 0.6 api hasn't really changed 2009 despite being "unstable". If changes were to happen we would likely find out before hand anyway.

pedrocamargo (approved):

Jake-Moss:
Merged into 2.0 cf76ef82

PR #818, opened 2026-07-30
Use total flow instead of first user class.

Jake-Moss:
Thanks Jan, thanks Pedro.


Jake-Moss:
Merged into 2.0 in 222b55c9

PR #828, opened 2026-08-11
New trigger to protect a_node and b_node

pedrocamargo:


.github/workflows/test_aequilibrae_environment.yml
```diff
@@ -32,5 +32,5 @@ jobs:
 
     - name: Runs test
       run: |
-        python -m uv pip install numpy==1.26.4 --force-reinstall
+        python -m uv pip install numpy==2.4.6 --force-reinstall
```
tomstephen, line 35: Do we need to be this prescriptive of the numpy version? Is >2 acceptable? Same in `pyproject.toml`. 
```suggestion
        python -m uv pip install numpy>2 --force-reinstall
```

pyproject.toml
```diff
@@ -28,7 +28,7 @@ requires-python = ">=3.11" # match classifiers above and ruff version below
 
 dependencies = [
   "geopandas",
-  "numpy",
+  "numpy>=2.4.6",
```
tomstephen, line 31: Are we using anything newer that what was introduced in Numpy 2?
```suggestion
  "numpy>2",
```

tests/aeq/project/test_network_triggers.py
```diff
@@ -68,3 +135,295 @@ def test_link_direction(sioux_falls_example):
...
+    _, replaced_node, _ = _build_adjacent_network(empty_project)
+
+    with empty_project.db_connection as conn:
+        conn.execute(
+            "INSERT INTO nodes (node_id, is_centroid, geometry) VALUES (9001, 1, GeomFromWKB(?, 4326))",
+            (Point(9, 9).wkb,),
+        )
+        before = _network_snapshot(conn)
+
+        with pytest.raises(sqlite3.IntegrityError):
```
tomstephen, line 377: Only `sqlite3.IntegrityError` without a `match`, intentional?

tomstephen (commented):
Looks good! Works locally and tests seem sufficient. Following up with a commit for some documentation.

.github/workflows/test_aequilibrae_environment.yml
pedrocamargo: 2.4.6 is the current version in QGIS, which is the only one we test other than the latest one, but I don't follow the NumPy development very closely.

Happy to go either way. @Jake-Moss

pyproject.toml
pedrocamargo: Same as above.

aequilibrae/project/database_specification/network/migrations/003_protect_link_end_nodes.py
```diff
@@ -0,0 +1,103 @@
...
+                SELECT count(*)
+                FROM nodes
+                WHERE nodes.geometry = {endpoint_function}(links.geometry)
+                  AND nodes.ROWID IN (
+                      SELECT ROWID
+                      FROM SpatialIndex
+                      WHERE f_table_name = 'nodes'
+                        AND search_frame = {endpoint_function}(links.geometry)))
+            """
+        )
```
Jake-Moss: Do we want to repair the a/b_node here or the geometry? a/b_node changes will change the topology of the network while changing the geometry will only change how the network is visualised.

If someone has modified the a/b_nodes previously and not updated the geometry, presumably the current a/b_nodes are correct.

If the geometry was updated (QGIS) then the triggers already handled the a/b_node updates.

To me this looks like it will break existing networks by unintentionally reverting changes.

Jake-Moss: Perhaps we also should add an easy-to-use backup method using sqlite backup methods.

Jake-Moss (approved):

PR #849, opened 2026-09-16
Turn penalties, BFW improvements, path finding result rework

Jake-Moss:
This PR includes a number of things for 2.0, and I wanted to get it merged before continuing on this branch.

This PR includes:
  - The initial turn penalties work by @pedrocamargo 
  - The improves of jzill/bfw to the line search methods, fixing a handful of issues and adding some nice improvements.
  - A rework of how we store path finding and related operations results, required information, and scratch buffers. This groups previously scatter bits of data into:
    -  Multiple "context" object, these hold static data that is required to perform an operation on something. For example it holds the forward start, costs (time) of each link, the link heads, and some counts. This is the only information required to conduct a path finding operation.
    - Multiple "result" objects, these hold the results of said operations. Each operation has it's own "result" that allocates the memory and such of the correct shape, and knows how to interrogate these results.
    - Two "query" objects, just for path finding and link loading, these contain extra information that may frequently change for an operation.
    - Multiple "workspace" objects, these a thread independent scratch buffers for each operation. They're designed to be re-used and combined.

The rework of result storage has *not* been integrated into the rest of the code base, instead it's been implemented in parallel to the existing code (hence the large +, but no - diff) but I'd like to get it and the turn penalties in as this will just continue to grow.

The result rework comes as an effort to unify the storage of node and turn base path finding results, in such a way that code consuming them does not have to be aware of the turn penalties.

Same content as #848  but squashed into smaller commits.


Jake-Moss:
Merging with the caveat that the new results objects are *not* integrated into the rest of code base. That will come later, with a large negative diff

